Issue #53, opened 2024-02-29
[Bug]: LDSC colinearity/unused column


sof202:
### What happened?

This particular line:
> 80 overlapany<-in.region(snpRegions, allReg)

in integrative/chromHMM/createAnnotationFiles.r

is unused in the output annotation file, making the line unused throughout the entire file.

This means that a previous bedr line:

> 52 allReg<-bedr.merge.region(bedr.sort.region(c(allRegions.atac, totalRegions.atac, neunRegions.atac, soxRegions.atac, allRegions.5hmc, totalRegions.5hmc, neunRegions.5hmc, soxRegions.5hmc,dnegRegions.5hmc)))

uses unnecessary computing power (as the result of this command is also unused in the output).

However, If the line was used in the annotation file, it is not necessarily non-colinear with the other columns being added to the annotation file. This is bad as collinearity in annotation columns will result in ldsc throwing an error (presumably due to some matrix manipulation in the calculation process).

Issue #46, opened 2024-02-21
[Enhancement suggestion]:  No config file template given for LDSC scripts


sof202:
### Enhancement suggestion

There is no configuration file template given for the script:
integrative/chromHMM/jobSubmission/runLDSCNeuralPeaks.sh

Adding one will assist users in correctly using this script.

Issue #45, opened 2024-02-21
[Code Refactor]: Out of date LDSC scripts


sof202:
### Path to affected file

```shell
integrative/chromHMM/jobSubmission/runPartionedHeritabilityOnPeaks.sh
integrative/chromHMM/processing/createAnnotationFiles.R
```

### Type of refactor

Edge case error handling, Code readability improvement

### Description of required code refactor

Multiple lines have unneeded semi colons, for example:
> 6  for chr in {1..22};

Line 16 is written incorrectly and is too specific to a past project
> 16 gwastraits=($ls ${LDREFPATH}/gwas_traits/Anorexia*.gz)

Should be:

> 16 gwastraits=$(ls ${LDREFPATH}/gwas_traits/Anorexia*.gz)

Also lots of lines are pointing towards specific old directories, this should be replaced with a configuration file.

The R script needs refactoring as it is very repetitive (needs some functions and rework)

Issue #44, opened 2024-02-21
[Bug]: Outdated LDSC scripts


sof202:
### What happened?

Affected file: integrative/chromHMM/jobSubmission/runLDSCNeuralPeaks.sh

Currently this file expects the user to run this script from the root of this repository and on line 43 we have:
> 43 source LDScoreRegression/runPartionedHeritabilityOnPeaks.sh

Which does not exist. 

further:
> 40 source activate ldsc

Is an old way of activating conda environments and should be changed to `conda activate path/to/env`. The addition of a file path here will reduce errors associated with different users using the script. The file path can be added into the configuration file.

PR #54, opened 2024-02-29
LDSC improvements

sof202:
# Description

This pull request will introduce a variety of improvements and fixes to the ldsc scripts previously located in the ChromHMM scripts directory. Highlights of these are as follows:
- Create a README file that provides the format expected for configuration files, the expected file structure and the usage of the scripts
- Removing repetition with functions and lists in the associated R script
- Removing unused lines in R script (overlap for peak called regions of all input peak files)
    - If this was actually implemented into the resulting annotation files, this runs the risk of introducing collinearity (which throws errors when inputted into ldsc)
- Refactor the main function to work with configuration files (which can be accessed via command line input instead of with full file path)
- Update log file management and SLURM directives to make script more versatile (no longer needs to be ran from a specific directory to work)
- Fix a variety of errors present in `runPartitionedHeritabilityOnPeaks.sh` including:
    - gwas traits not being initialised properly (due to incorrect placement of $ in command)
    - Remove any errors that could occur due to word splitting or globbing (https://www.shellcheck.net/wiki/SC2068)
    - Removed old repeated code that would run for multiple analyses on the same dataset (this is not required anymore as the script is more modular now)
- Renamed scripts (correcting spelling errors)
- Moved scripts to their own directory, ldsc may be related to chromHMM, but it is not dependent on chromHMM (and vice versa). The scripts are isolated to reflect this.

## Issue ticket number

This pull request is to address issues: #44, #45, #46, #53.

## Type of pull request

- [ ] Bug fix
- [ ] New feature/enhancement
- [x] Code refactor
- [ ] Documentation update

## Checklist


integrative/LDSC/README.md
ejh243, line 1: Do we need a general comment that this script is designed to take a specific set of epigenetic annotations we have generated and therefore currently isn't very generalisable.

sof202: I've added to the README.md what kinds of data can be used. Technically all the script requires is bed files. But I've specified peak-called data to be consistent with the name of the script

ejh243 (approved):

PR #65, opened 2024-03-08
create new initiateDataFolders.sh script

NiCl2:
# Description

This pull request will create a new bash script in general/jobSubmissions for initiating the default data folders

## Issue ticket number

This pull request is to address issue: #7 .

## Type of pull request

- [ ] Bug fix
- [x] New feature/enhancement
- [ ] Code refactor
- [ ] Documentation update

## Checklist

### DIFF
--- a/general/jobSubmission/initiateDataFolders.sh
+++ b/general/jobSubmission/initiateDataFolders.sh
@@ -5,7 +5,7 @@ set -e
 
 # one argument: directories location
 
-cd $1
+cd "$1"
 
 mkdir 0_metadata 1_raw 2_trimmed 3_aligned 4_catMemes
 
--- a/general/jobSubmission/initiateDataFolders.sh
+++ b/general/jobSubmission/initiateDataFolders.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+# create default data directories in given location
+
+set -e
+
+# one argument: directories location
+
+cd $1
+
+mkdir 0_metadata 1_raw 2_trimmed 3_aligned 4_catMemes
+
+


general/jobSubmission/initiateDataFolders.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# create default data directories in given location
+
+set -e
+
+# one argument: directories location
+
+cd $1
```
sof202, line 8: Minor: Could you add quotations around $1 to prevent splitting. This command wouldn't fail, but give undefined behaviour if a folder is entered that has a space in it.

general/jobSubmission/initiateDataFolders.sh
```diff
@@ -0,0 +1,12 @@
...
+
+mkdir 0_metadata 1_raw 2_trimmed 3_aligned 4_catMemes
```
sof202, line 10: This is great, I like how it is all on one line, efficient. I forgot you could do that.

general/jobSubmission/initiateDataFolders.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
sof202, line 1: This file is in "jobSubmission" which might imply that the script it to be submitted via `sbatch`. Do you think this is a problem? This script clearly doesn't need to be submitted via `sbatch`, its basically a one liner.

sof202 (approved):
Fantastic, thank you for your contribution

general/jobSubmission/initiateDataFolders.sh
NiCl2: ye I wasn't sure where to put it

PR #64, opened 2024-03-08
fix functionsR load

SiyiSEA:
# Description

This pull request will fix the path for direct to the funtionsR folder.

## Issue ticket number

This pull request is to address issue: #60 .

## Type of pull request

- [x] Bug fix
- [ ] New feature/enhancement
- [ ] Code refactor
- [ ] Documentation update

## Checklist

- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have tested my code to check that it is functional
- [ ] I have used linters to check for common sources of errors
- [ ] I have implemented fail safes in my code to account for edge cases
- [ ] I have made the corresponding changes to the documentation

PR #69, opened 2024-03-08
Revert "fix functionsR load"

sof202:
Reverts ejh243/BrainFANS#64

The proposed changes do not actually fix the problem. The files still won't load if BrainFANS is not located in the user's home directory.


SiyiSEA (approved):
revert

Issue #60, opened 2024-03-06
[Bug]: load R functions package command does not work


ejh243:
### Contact Details

e.j.hannon@exeter.ac.uk

### What happened?

When loading internal R library in [load GDS script](https://github.com/ejh243/BrainFANS/blob/master/array/DNAm/preprocessing/loadDataGDS.r) on line 37  (and possibly others) relative path no longer works. 

### How can the bug be reproduced?

`devtools::load_all(path = "../functionsR")`

ejh243:
Also a question as to whether this package should be external to the pipeline - I think possibly so...

SiyiSEA:
fixed

PR #70, opened 2024-03-08
fix the path to funtionsR

SiyiSEA:
# Description

This pull request will fix the path to the functions R.
We add another vairable named RPATH to let the user specify the path to the R function in the config file.
And we add some comments in the top of the BrainFANS/array/DNAm/preprocessing/jobSubmission/1_runDNAmQC.sh, which describe the inputs.

## Issue ticket number

This pull request is to address issue: #60 .

## Type of pull request

- [x] Bug fix
- [ ] New feature/enhancement
- [ ] Code refactor
- [ ] Documentation update

## Checklist

### DIFF
--- a/array/DNAm/preprocessing/jobSubmission/1_runDNAmQC.sh
+++ b/array/DNAm/preprocessing/jobSubmission/1_runDNAmQC.sh
@@ -10,11 +10,20 @@
 #SBATCH --output=DNAm/logFiles/%u/QCDNAdata.log # output file
 #SBATCH --job-name=QCDNAdata
 
+#------------------------------------------------------
+
+# 1. the first input is your project folder.
+# 2. the second input is the full path to config file.
+# 3. the third input is a the path to the config.r file.
+
+#-----------------------------------------------------
+
+
 ## print start date and time
 echo Job started on:
 date -u
 JOBNAME="QCDNAdata"
-    
+
 ## needs to be executed from the scripts folder
 echo "Changing Folder to: " $SLURM_SUBMIT_DIR
 cd $SLURM_SUBMIT_DIR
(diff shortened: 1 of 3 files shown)


SiyiSEA:
I did the merge confict! done!👍

array/DNAm/preprocessing/jobSubmission/1_runDNAmQC.sh
```diff
@@ -10,33 +10,35 @@
 #SBATCH --output=DNAm/logFiles/%u/QCDNAdata.log # output file
 #SBATCH --job-name=QCDNAdata
 
+#------------------------------------------------------
+
+# 1. the first input is your project folder.
+# 2. the second input is the full path to config file.
+# 3. the third input is a the path to the config.r file.
+
+#-----------------------------------------------------
```
sof202, line 19: Thanks for putting this at the top of the page. It can be quite difficult for the user sometimes to know what the inputs to the script does. This makes it much more transparent.

array/DNAm/preprocessing/jobSubmission/1_runDNAmQC.sh
```diff
@@ -10,33 +10,35 @@
...
-## default file used unless alternative provided on command line
-if [ -z "$2" ];
-then
-    RCONFIG=$SLURM_SUBMIT_DIR/DNAm/config/config.r
-else
-    RCONFIG=$2
-fi
+# the second input should be config file
+source $2 || exit 1
+RCONFIG=$3
```
sof202, line 36: This is a much better way of doing it than before. now the config files (for the shell script and the Rscripts) can be placed anywhere. The user specifying the path means we don't need to force the script from being ran in a particular spot anymore.

array/DNAm/preprocessing/jobSubmission/1_runDNAmQC.sh
```diff
@@ -10,33 +10,38 @@
...
+
+
 ## print start date and time
 echo Job started on:
 date -u
 JOBNAME="QCDNAdata"
-    
+
+# Forcibly moves the user to the scripts directory
+cd $(dirname $0)/../../../..
```
sof202, line 28: Although this looks a bit ugly, I can't think of a better way of doing this. It will have to do. I guess another way of doing this would be to ask the user to put the location of the repository into the config file?

sof202 (approved):
These are good changes, the code has improved.
Thanks for your contributions!

PR #71, opened 2024-03-13
Script to set up for ATAC-seq pipeline

marinafloresp:
# Description

This pull request will add the updated scripts of the first and second steps of the ATACseq pipeline, as well as the script to set up the pre-requisites to run the pipeline.

## Issue ticket number

This pull request is to address issue: #[72]. A script to check pre-requisites and the required set-up to run the ATAC seq pipeline has been created.

## Type of pull request

- [ ] Bug fix
- [X] New feature/enhancement
- [X] Code refactor
- [ ] Documentation update

## Checklist

### DIFF
--- a/sequencing/ATACSeq/jobSubmission/0_setUp.sh
+++ b/sequencing/ATACSeq/jobSubmission/0_setUp.sh
@@ -1,0 +1,174 @@
+#!/bin/bash
+#SBATCH --export=ALL # export all environment variables to the batch job.
+#SBATCH -p mrcq # submit to the serial queue
+#SBATCH --time=48:00:00 # Maximum wall time for the job.
+#SBATCH -A Research_Project-MRC190311 # research project to submit under. 
+#SBATCH --nodes=1 # specify number of nodes.
+#SBATCH --ntasks-per-node=16 # specify number of processors per node
+#SBATCH --mail-type=END # send email at job completion 
+#SBATCH --output=ATACSetUpS0-%A.log
+#SBATCH --error=ATACSetUpS0-%A.err
+#SBATCH --job-name=ATACSetUpS0
+
+## ==========================================================================================================##
+##                               ATAC-seq pipeline STEP 0: Set up                                            ##
+## ==========================================================================================================##
+## EXECUTION: sbatch ./sequencing/ATACSeq/jobSubmission/0_setUp.sh <project name>                            ||
+## - execute from scripts directory                                                                          ||
+##                                                                                                           ||
+## INPUTS:                                                                                                   || 
+## $1 -> <project name> name of project to set up pipeline for.                                              || 
+##                                                                                                           ||
+## DESCRIPTION: This script sets up the required directories, libraries and modules to run the ATAC-seq      || 
+## pipeline. Will check the config file for the project and the initial paths required for the first steps of|| 
+## the analysis.                                                                                             ||
+##                                                                                                           ||
+## REQUIRES:                                                                                                 ||
+## - File in ${METADIR}/samples.txt that lists sample names.                                                 ||
+## ==========================================================================================================##
+
+
+## ============ ##
+##    SET UP    ##
+## ============ ##
+
+## print start date and time
+echo Job started on:
+date -u
+
+
+LOG_DIR=ATACSeq/logFiles/${USER}/${SLURM_JOB_ID}
+echo "Log files will be moved to dir: " $LOG_DIR
+mkdir -p $LOG_DIR
+mv ATACSetUpS0-${SLURM_JOB_ID}* $LOG_DIR
+
+## ============ ##
+##    CHECKS    ##
+## ============ ##
+
+## CONFIG FILE ##
+
+## Check if config file for input project exist
+
+if [[ -s "/lustre/projects/Research_Project-MRC190311/ATACSeq/$1/config.txt" ]]
+then
+  source "/lustre/projects/Research_Project-MRC190311/ATACSeq/$1/config.txt"
+  echo "Config file for $1 project found."
+else 
+  { echo "Config file for chosen project does not exist or is in the wrong directory. Please create this first." ; exit 1; }  
+fi
+
+## Check if R config file for input project exist
+
+if [[ -s "/lustre/projects/Research_Project-MRC190311/ATACSeq/$1/config.r" ]]
+then
+  echo "R config file for $1 project found."
+else 
+  { echo "R config file for chosen project does not exist or is in the wrong directory. Please create this first." ; exit 1; }  
+fi
+
+## DIRECTORIES ##
+
+## Check that the metadata directory exists and contains sample list.
+if [[ ! -d ${METADIR} ]]
+then
+	echo 'METADATA directory does not exist'
+	exit 1
+elif  test -f ${METADIR}/samples.txt; then
+	mapfile -t SAMPLEIDS < ${METADIR}/samples.txt 
+	echo "Sample list found. Number of sample IDs found:" """${#SAMPLEIDS[@]}"""
+else
+	echo 'samples.txt file does not exist in METADIR'
+fi
+
+## Check that the sampleSheet exist
+if [[ -d ${METADIR} ]] && [[ -s "${METADIR}/sampleSheet.csv" ]]
+then
+  SAMPLEIDS2=( $(tail -n +2 ${METADIR}/sampleSheet.csv  | cut -d ',' -f2) )
+  echo "Sample sheet found. Number of sample IDs found:"" ""${#SAMPLEIDS2[@]}"""
+else
+  echo "sampleSheet.csv file does not exist in METADIR."
+fi
+
+## Check if remaining main directories exist and create if not
+dirs=(${RAWDATADIR} ${FASTQCDIR} ${TRIMDIR} ${ALIGNEDDIR} ${PEAKDIR})
+type=("DATA" "FASTQC" "TRIMMED" "ALIGNED" "PEAK CALLED")
+
+for (( x=0; x<${#dirs[@]}; x++ ))
+do
+	if [ -d ${dirs[x]} ] 
+	then 
+		echo ${type[x]} "directory:" ${dirs[x]}
+	elif [ ${dirs[x]} = $RAWDATADIR ]; then 
+		echo ${dirs[x]} "directory does not exist"
+		exit 1
+	else
+		echo "Creating directory: " ${dirs[x]}
+		mkdir -p ${dirs[x]} 
+	fi
+done
+
+## MODULES ##
+
+## Check that the SAMstats environment exists and create it if not
+
+module load Anaconda3
+if [ -d ${ENVDIR} ]
+then 
+	echo "ENCODEQC environment exists"
+  source $CONDAENV
+  conda activate ${ENVDIR}
+  echo "List of packages included in conda environment is output in project directory"
+  conda list > "/lustre/projects/Research_Project-MRC190311/ATACSeq/$1/condaEnv.txt"
+  conda deactivate
+else
+	echo "ENCODEQC environment does not exist"
+	echo "Creating environment"
+	conda create --prefix ${ENVDIR}
+  source $CONDAENV
+	conda activate ${ENVDIR}
+	echo "Installing SAMstats"
+	conda install -c bioconda samstats
+	conda deactivate
+	echo "ENCODEQC environment has been created and SAMstats installed"
+fi
+
+## Check that pip environment exist and create it if not
+module load Python/3.9.6-GCCcore-11.2.0-bare
+
+if [[ -d ${PIPENV} ]]
+then
+  echo "Pip environment exists."
+  source /lustre/home/${USER}/pythonEnv/bin/activate
+  echo "List of packages included in pip environment is output in project directory"
+  pip list > "/lustre/projects/Research_Project-MRC190311/ATACSeq/$1/pipEnv.txt"
+  deactivate
+else
+  echo "Pip environment does not exist."
+  echo "Creating environment."
+  python -m venv ${PIPENV}
+  source /lustre/home/${USER}/pythonEnv/bin/activate
+  echo "Installing basic libraries needed in pipeline"
+  pip install Cython
+  pip install cykhash
+  pip install numpy
+  pip install hmmlearn
+  pip install scikit-learn
+  pip install scipy
+  pip install macs3
+  deactivate
+fi
+
+module purge
+
+echo "R v4.2 is required and will be loaded if found"
+## Check that required R version 4.2 exist
+module load R/4.2.1-foss-2022a
+
+echo "R packages will be checked and installed if not found."
+## Check that required R libraries are installed
+Rscript ${SCRIPTDIR}/ATACSeq/preprocessing/setUpRenv.r ${USER}
+
+echo "Set up for project $1 complete."
+  
+  
--- a/sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
+++ b/sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #SBATCH --export=ALL # export all environment variables to the batch job.
 #SBATCH -p mrcq # submit to the serial queue
 #SBATCH --time=48:00:00 # Maximum wall time for the job.
@@ -6,8 +6,8 @@
 #SBATCH --nodes=1 # specify number of nodes.
 #SBATCH --ntasks-per-node=16 # specify number of processors per node
 #SBATCH --mail-type=END # send email at job completion 
-#SBATCH --output=/lustre/projects/Research_Project-MRC190311/scripts/sequencing/ATACSeq/logFiles/%u/ATACpreAnalysisS1-%A_%a.o
-#SBATCH --error=/lustre/projects/Research_Project-MRC190311/scripts/sequencing/ATACSeq/logFiles/%u/ATACpreAnalysisS1-%A_%a.e
+#SBATCH --output=/lustre/projects/Research_Project-MRC190311/scripts/sequencing/ATACSeq/logFiles/%u/ATACpreAnalysisS1-%A_%a.log
+#SBATCH --error=/lustre/projects/Research_Project-MRC190311/scripts/sequencing/ATACSeq/logFiles/%u/ATACpreAnalysisS1-%A_%a.err
 #SBATCH --job-name=ATACpreAnalysisS1
 
 ## ===================================================================================================================##
@@ -17,12 +17,12 @@
 ## CONTACT: m.flores-payan@exeter.ac.uk                                                                               ||
 ## LAST UPDATE: February 2024                                                                                         ||
 ## ===================================================================================================================##
-## EXECUTION: sbatch --array= ./sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh <projectName> <option>      ||
+## EXECUTION: sbatch --array= ./sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh <config file> <option>      ||
 ## - execute from scripts directory                                                                                   ||
 ##                                                                                                                    ||
 ## INPUTS:                                                                                                            || 
 ## --array -> Number of jobs to run. Will select sample(s) corresponding to the number(s) input                       ||
-## $1 -> <projectName> Name of project                                                                                ||
+## $1 -> <config file> directory of config file for project analysed
 ## $2 -> <option> Specify step to run: FASTQC, TRIM, ALIGN or ENCODE. Can be combined. Default is to                  ||
 ## run all                                                                                                            ||
 ##                                                                                                                    ||
@@ -42,14 +42,9 @@
 echo Job started on:
 date -u
 
-## needs to be executed from the scripts folder
-echo "Changing Folder to: " $SLURM_SUBMIT_DIR
-cd $SLURM_SUBMIT_DIR || exit 1
-
 ## load config file provided on command line related to the specified project
-echo "Loading config file for project: " $1
-export PROJECT=$1
-source ./ATACSeq/config/config.txt 
+source $1
+echo "Loading config file for project: " ${PROJECT}
 
 ## check directories
 echo "Project directory is: " $DATADIR 
@@ -80,9 +75,8 @@ then
     
     ## Sample(s) specified in by array number(s) from command line
     sampleID=${SAMPLEIDS[${SLURM_ARRAY_TASK_ID}]}
-    dirs=( ${RAWDATADIR} ${RAWDATADIR_B1} ${RAWDATADIR_B2} )
-    #toProcess=($(find ${RAWDATADIR} -maxdepth 1 -name ${SAMPLEIDS[${SLURM_ARRAY_TASK_ID}]}'*'))
-    toProcess=($(find "${dirs[@]}" -maxdepth 1 -name ${SAMPLEIDS[${SLURM_ARRAY_TASK_ID}]}'*'))
+    toProcess=($(find ${RAWDATADIR} -maxdepth 1 -name ${SAMPLEIDS[${SLURM_ARRAY_TASK_ID}]}'*'))
+    
     ## sort the toProcess array so that R1 and R2 are consecutive 
     IFS=$'\n' # need to set this as \n rather than default - a space, \t and then \n - so that elements are expanded using \n as delimiter
     toProcess=($(sort <<<"${toProcess[*]}")) ## sort so that the first element is R1
@@ -95,8 +89,7 @@ then
     ## If only $1 is specified, all steps will be run
     ## option FASTQC: run sequencing QC on raw reads files        
     if [ $# == 1 ] || [[ $2 =~ 'FASTQC' ]]
-    then
-      
+    then      
         echo "Running step 1.1 of ATAC-seq pipeline: FASTQC"
         module load FastQC 
         cd ${SCRIPTDIR}
@@ -139,21 +132,21 @@ then
           
       ## load conda env for samstats
       module load Anaconda3
-      source activate encodeqc
+      source $SOURCE
+      conda activate encodeqc
 
       cd ${SCRIPTDIR}
       sh ./ATACSeq/preprocessing/calcENCODEQCMetrics.sh ${sampleID}
     fi
   
     ## move log files into a folder when process(es) completed
-    cd ATACSeq/logFiles/${USER}/
-    mkdir -p ${SLURM_ARRAY_JOB_ID}
-    mv *${SLURM_ARRAY_JOB_ID}_${SLURM_ARRAY_TASK_ID}.* ${SLURM_ARRAY_JOB_ID}
-
+    
+    LOG_DIR=ATACSeq/logFiles/${USER}/${SLURM_ARRAY_JOB_ID}
+    mkdir -p $LOG_DIR
+    mv "./ATACSeq/logFiles/${USER}/ATACpreAnalysisS1${SLURM_ARRAY_JOB_ID}_${SLURM_ARRAY_TASK_ID}*" $LOG_DIR
 
 else
     echo "File list not found"
 fi
 
-echo 'EXIT CODE: ' $?
 
--- a/sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
+++ b/sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
@@ -22,7 +22,7 @@
 ##                                                                                                                    ||
 ## INPUTS:                                                                                                            || 
 ## --array -> Number of jobs to run. Will select sample(s) corresponding to the number(s) input                       ||
-## $1 -> <config file> directory of config file for project analysed
+## $1 -> <config file> directory of config file for project analysed                                                  ||
 ## $2 -> <option> Specify step to run: FASTQC, TRIM, ALIGN or ENCODE. Can be combined. Default is to                  ||
 ## run all                                                                                                            ||
 ##                                                                                                                    ||
@@ -48,7 +48,14 @@ echo "Loading config file for project: " ${PROJECT}
 
 ## check directories
 echo "Project directory is: " $DATADIR 
-echo 'Script is running from directory: ' ${SCRIPTDIR}
+echo "Script is running from directory: " ${SCRIPTDIR}
+
+## Log files directory
+LOG_DIR=ATACSeq/logFiles/${USER}/${SLURM_ARRAY_JOB_ID}
+
+echo "Log files will be moved to dir: " $LOG_DIR
+mkdir -p $LOG_DIR
+mv "./ATACSeq/logFiles/${USER}/ATACpreAnalysisS1${SLURM_ARRAY_JOB_ID}_${SLURM_ARRAY_TASK_ID}*" $LOG_DIR
 
 ##check array specified and exit if not
 if [[ ${SLURM_ARRAY_TASK_ID} == '' ]]
@@ -82,9 +89,8 @@ then
     toProcess=($(sort <<<"${toProcess[*]}")) ## sort so that the first element is R1
     unset IFS 
     echo ${toProcess}
-    echo "R1 file found is: " $( basename ${toProcess[0]} )
-    
-    echo "Current sample: " ${sampleID} 
+    echo "R1 file found is: " $( basename ${toProcess[0]} ) 
+    echo "Sample to be used is: " ${sampleID} 
     
     ## If only $1 is specified, all steps will be run
     ## option FASTQC: run sequencing QC on raw reads files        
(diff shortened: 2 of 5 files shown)


sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
```diff
@@ -0,0 +1,159 @@
...
+##    SET UP    ##
+## ============ ##
+
+## print start date and time when script runs
+echo Job started on:
+date -u
+
+## needs to be executed from the scripts folder
+echo "Changing Folder to: " $SLURM_SUBMIT_DIR
+cd $SLURM_SUBMIT_DIR || exit 1
```
sof202, line 47: These commands are not doing anything, I don't know why they were in the original file.

ejh243:
They are a legacy from when I was struggling to work out why my script didn't run. Can be deleted.

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
```diff
@@ -0,0 +1,159 @@
...
+
+## load config file provided on command line related to the specified project
+echo "Loading config file for project: " $1
+export PROJECT=$1
+source ./ATACSeq/config/config.txt 
+
```
sof202, line 53: If I remember correctly, $PROJECT is used in the config so that the config file can be more modular and work with multiple projects.
In a later pull request (doesn't need to be this one), it would be nice if the config file is one of the inputs from the user instead of $PROJECT. Putting the config file as a user input will also mean that the script doesn't need to be executed from the root of the repository.
This would require the config file to be specific to a project (though this is what was pushed for a couple of weeks ago), and $PROJECT wouldn't be required anymore.

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
```diff
@@ -0,0 +1,159 @@
...
+## check directories
+echo "Project directory is: " $DATADIR 
+echo 'Script is running from directory: ' ${SCRIPTDIR}
+
+##check array specified and exit if not
+if [[ ${SLURM_ARRAY_TASK_ID} == '' ]]
+then 
+    { echo "Job does not appear to be an array. Please specify --array on the command line." ; exit 1; }
+fi
```
sof202, line 62: Instead of checking if the array is submitted as an array. We could add a SLURM directive at the top of the script:
`#SBATCH --array=1-4` (or similar)

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
```diff
@@ -0,0 +1,159 @@
+#!/bin/sh
```
sof202, line 1: Could you change this to use bash instead of sh?
Just for consistency across the repository

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
sof202: Nevermind, I've read further and it seems that the user should specify this themselves as otherwise not all files will be processed.
I guess in the future this could be improved by introducing loops and splitting the samples up into each of the jobs in the array.

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
```diff
@@ -0,0 +1,159 @@
...
+    echo "Number of sample IDs found:"" ""${#SAMPLEIDS[@]}"""
+    
+    ## Sample(s) specified in by array number(s) from command line
+    sampleID=${SAMPLEIDS[${SLURM_ARRAY_TASK_ID}]}
+    dirs=( ${RAWDATADIR} ${RAWDATADIR_B1} ${RAWDATADIR_B2} )
+    #toProcess=($(find ${RAWDATADIR} -maxdepth 1 -name ${SAMPLEIDS[${SLURM_ARRAY_TASK_ID}]}'*'))
+    toProcess=($(find "${dirs[@]}" -maxdepth 1 -name ${SAMPLEIDS[${SLURM_ARRAY_TASK_ID}]}'*'))
+    ## sort the toProcess array so that R1 and R2 are consecutive 
+    IFS=$'\n' # need to set this as \n rather than default - a space, \t and then \n - so that elements are expanded using \n as delimiter
+    toProcess=($(sort <<<"${toProcess[*]}")) ## sort so that the first element is R1
```
sof202, line 88: I'm not convinced this is actually sorting the array. This command is parsing the array as a string and so the sort will just output the array as is.

Observe:
```bash
[me@login01 test] $ array=(apple pear banana orange)
[me@login01 test] $ sort <<<"${array[*]}"
apple pear banana orange
```

I found this solution online if this actually matters:
`toProcess=($(printf "%s\n" "${toProcess[@]}" | sort))`
I can't figure out how to get this to work with a regular `echo` however.

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
```diff
@@ -0,0 +1,159 @@
...
+then 
+    ## create an array from the file
+    mapfile -t SAMPLEIDS < ${METADIR}/samples.txt 
+    echo "Number of sample IDs found:"" ""${#SAMPLEIDS[@]}"""
+    
+    ## Sample(s) specified in by array number(s) from command line
+    sampleID=${SAMPLEIDS[${SLURM_ARRAY_TASK_ID}]}
+    dirs=( ${RAWDATADIR} ${RAWDATADIR_B1} ${RAWDATADIR_B2} )
+    #toProcess=($(find ${RAWDATADIR} -maxdepth 1 -name ${SAMPLEIDS[${SLURM_ARRAY_TASK_ID}]}'*'))
+    toProcess=($(find "${dirs[@]}" -maxdepth 1 -name ${SAMPLEIDS[${SLURM_ARRAY_TASK_ID}]}'*'))
```
sof202, line 85: This works great!
I can only see this breaking if any of the directories have spaces in them (hopefully they don't)

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
```diff
@@ -0,0 +1,159 @@
...
+    	  echo "Running step 1.2 of ATAC-seq pipeline: TRIM"
+        cd ${SCRIPTDIR}
+        sh ./preScripts/fastp.sh ${sampleID} ${toProcess[0]} ${toProcess[1]} 
+    fi
+
+    ## option ALIGN: run alignment and post processing on sample
+	  if [ $# == 1 ] || [[ $2 =~ 'ALIGN' ]]
+  	then
+   
+  		echo "Running step 1.3 of ATAC-seq pipeline: ALIGN"
```
sof202, line 121: I like the transparency of adding in these logging messages.

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
```diff
@@ -0,0 +1,159 @@
...
+  	if [ $# == 1 ] || [[ $2 =~ 'ENCODE' ]]
+  	then
+      echo "Running step 1.4 of ATAC-seq pipeline: ENCODE"
+  		module purge
+  		module load SAMtools
+  		module load BEDTools/2.27.1-foss-2018b ##necessary to specify earlier BEDTools version
+          
+      ## load conda env for samstats
+      module load Anaconda3
+      source activate encodeqc
```
sof202, line 142: Would you mind updating this to `conda activate`?
`source activate` is the old way of activating a conda environment and is generally slower.
Also, is there any explanation of how to setup this conda environment in associated documentation?

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
```diff
@@ -0,0 +1,159 @@
...
+    cd ATACSeq/logFiles/${USER}/
+    mkdir -p ${SLURM_ARRAY_JOB_ID}
+    mv *${SLURM_ARRAY_JOB_ID}_${SLURM_ARRAY_TASK_ID}.* ${SLURM_ARRAY_JOB_ID}
+
+
+else
+    echo "File list not found"
+fi
+
+echo 'EXIT CODE: ' $?
```
sof202, line 158: All this is doing is checking if the exit status of the last command. The options for that are the:
`echo "File list not found"` line
and the
`mv *${SLURM_ARRAY_JOB_ID}_${SLURM_ARRAY_TASK_ID}.* ${SLURM_ARRAY_JOB_ID}`

Which means it isn't really doing much.

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
```diff
@@ -0,0 +1,159 @@
...
+## $2 -> <option> Specify step to run: FASTQC, TRIM, ALIGN or ENCODE. Can be combined. Default is to                  ||
+## run all                                                                                                            ||
+##                                                                                                                    ||
+## DESCRIPTION: This script performs the first step in the ATAC-seq pipeline: pre-analysis. This                      ||
+## includes quality control of raw reads (FASTQC), trimming the reads (TRIM), aligning reads to                       ||
+## reference genome (ALIGN) and calculate ENCODEQC metrics (ENCODE).                                                  ||
+##                                                                                                                    ||
+## REQUIRES:                                                                                                          ||
+## - File in ${METADIR}/samples.txt that lists sample names.                                                          ||
```
sof202, line 34: This information is appreciated, thank you

sof202 (approved):
The script is now better explained and will work with the different raw directories in the pipeline. Thank you

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
marinafloresp: This was a temporary fix as samples were in different subfolders in the 1_raw directory

marinafloresp (commented):
I have reviewed the comments on the script and will submit an updated version with the changes suggested

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
sof202: Does the script still work if you remove `-maxdepth 1`?
Will it pick up unwanted files that are deeper in the file system if it is removed?

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
```diff
@@ -0,0 +1,152 @@
...
+#SBATCH --output=/lustre/projects/Research_Project-MRC190311/scripts/sequencing/ATACSeq/logFiles/%u/ATACpreAnalysisS1-%A_%a.log
+#SBATCH --error=/lustre/projects/Research_Project-MRC190311/scripts/sequencing/ATACSeq/logFiles/%u/ATACpreAnalysisS1-%A_%a.err
+#SBATCH --job-name=ATACpreAnalysisS1
+
+## ===================================================================================================================##
+##                               ATAC-seq pipeline STEP 1: Pre-analysis                                               ##
+## ===================================================================================================================##
+## PROOFREADER : Marina Flores-Payan                                                                                  ||
+## CONTACT: m.flores-payan@exeter.ac.uk                                                                               ||
+## LAST UPDATE: February 2024                                                                                         ||
```
ejh243, line 18: As git logs this information we don't strictly need this info.

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
```diff
@@ -0,0 +1,152 @@
...
+  	then
+      echo "Running step 1.4 of ATAC-seq pipeline: ENCODE"
+  		module purge
+  		module load SAMtools
+  		module load BEDTools/2.27.1-foss-2018b ##necessary to specify earlier BEDTools version
+          
+      ## load conda env for samstats
+      module load Anaconda3
+      source $SOURCE
+      conda activate encodeqc
```
ejh243, line 136: Do we have somewhere a log of what is included in this conda env?

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
```diff
@@ -0,0 +1,152 @@
...
+
+      cd ${SCRIPTDIR}
+      sh ./ATACSeq/preprocessing/calcENCODEQCMetrics.sh ${sampleID}
+    fi
+  
+    ## move log files into a folder when process(es) completed
+    
+    LOG_DIR=ATACSeq/logFiles/${USER}/${SLURM_ARRAY_JOB_ID}
```
ejh243, line 144: Can we move the definition of this variable to the beginning of the script. Have all definitions together?

ejh243 (changes requested):
Move the variable definitions to the top of the script in the SET UP section.

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
```diff
@@ -42,14 +42,9 @@
...
 
-## needs to be executed from the scripts folder
-echo "Changing Folder to: " $SLURM_SUBMIT_DIR
-cd $SLURM_SUBMIT_DIR || exit 1
-
 ## load config file provided on command line related to the specified project
-echo "Loading config file for project: " $1
-export PROJECT=$1
-source ./ATACSeq/config/config.txt 
+source $1
```
sof202, line 46: Great, thank you

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
```diff
@@ -139,21 +132,21 @@ then
           
       ## load conda env for samstats
       module load Anaconda3
-      source activate encodeqc
+      source $SOURCE
+      conda activate encodeqc
```
sof202, line 136: Great, this is now more up to date

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
marinafloresp: I can get a list of the packages in the conda env

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
```diff
@@ -139,21 +132,21 @@ then
...
   
     ## move log files into a folder when process(es) completed
-    cd ATACSeq/logFiles/${USER}/
-    mkdir -p ${SLURM_ARRAY_JOB_ID}
-    mv *${SLURM_ARRAY_JOB_ID}_${SLURM_ARRAY_TASK_ID}.* ${SLURM_ARRAY_JOB_ID}
-
+    
+    LOG_DIR=ATACSeq/logFiles/${USER}/${SLURM_ARRAY_JOB_ID}
+    mkdir -p $LOG_DIR
+    mv "./ATACSeq/logFiles/${USER}/ATACpreAnalysisS1${SLURM_ARRAY_JOB_ID}_${SLURM_ARRAY_TASK_ID}*" $LOG_DIR
```
sof202, line 146: Love it, much more robust. The only problem would come from the user not setting up their log file directory (as this directory is user specific). If you were to remove the `%u` from the `#SBATCH` line (and the first reference to ${USER} on line 146) this would no longer be an issue. 
Also, from my testing, you can move the log and error files at the top of the script and it works just the same. This means that there won't be a bunch of temporary log files flooding the logFiles directory (if the script takes a long time to complete for example)

sof202 (approved):

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
marinafloresp: Okay, I thought that would happen only when the process is finished, although the files are created at the beginning, so it makes sense

sof202: Yeah, I have no idea why it works. I was using hard links before, but then I realised that just moving the files also works. I have no idea how SLURM is keeping track of where the log files are and how it is printing to them.

sequencing/ATACSeq/jobSubmission/2_batchCalcQCMetrics.sh
```diff
@@ -6,46 +6,68 @@
...
+#SBATCH --error=ATACSeq/logFiles/%u/ATACcalcQCS2-%A_%a.err
+#SBATCH --job-name=ATACcalcQCS2
 
-#-----------------------------------------------------------------------#
+## ===================================================================================================================##
+##                             ATAC-seq pipeline STEP 2: Post alignment processing                                    ##
+## ===================================================================================================================##
+## PROOFREADER : Marina Flores-Payan                                                                                  ||
+## CONTACT: m.flores-payan@exeter.ac.uk                                                                               ||
+## LAST UPDATE: February 2024                                                                                         ||
```
sof202, line 18: As per @ejh243 last comment, we expect to remove these lines as they can be found through git blame *etc.*

sequencing/ATACSeq/jobSubmission/2_batchCalcQCMetrics.sh
```diff
@@ -6,46 +6,68 @@
...
+##                                                                                                                    ||
+## INPUTS:                                                                                                            || 
+## --array -> It splits the samples into groups of 10 to run in parallel, so batch number should be the number of     ||
+##  samples divided by 10. e.g. If there are 50 samples, batch number should be 0-5, producing 5 batches of 10 samples|| 
+##  each. If the number of samples is less than 10, set batch number to 0                                             ||
+## - $1 -> <config file> : directory to config file for project                                                       ||
+##                                                                                                                    ||
+## DESCRIPTION: This script uses the ATACseqQC R package to generate the fragment distribution and calculate some     ||
+## summary statistics.                                                                                                ||
```
sof202, line 30: Could you specify what the summary statistics are?

sequencing/ATACSeq/jobSubmission/2_batchCalcQCMetrics.sh
```diff
@@ -6,46 +6,68 @@
...
-export PROJECT=$1
-source ./ATACSeq/config/config.txt 
-
 echo "Project directory is: " $DATADIR
 
+## Log files directory
+LOG_DIR=ATACSeq/logFiles/${USER}/${SLURM_ARRAY_JOB_ID}
+echo "Log files will be moved to dir: " $LOG_DIR
+mkdir -p $LOG_DIR
+mv "./ATACSeq/logFiles/${USER}/ATACcalcQCS2-${SLURM_ARRAY_JOB_ID}_${SLURM_ARRAY_TASK_ID}*" $LOG_DIR
```
sof202, line 52: Nice. Hopefully we can standardise how we treat log files in the future

sequencing/ATACSeq/jobSubmission/2_batchCalcQCMetrics.sh
```diff
@@ -6,46 +6,68 @@
...
+##  ==================  ##
 
 mkdir -p ${ALIGNEDDIR}/QCOutput
 
-module load R/3.6.3-foss-2020a
-
-Rscript ${SCRIPTDIR}/ATACSeq/preprocessing/fragmentDistribution.r ${ALIGNEDDIR} ${SLURM_ARRAY_TASK_ID} 
-
-echo 'EXITCODE: ' $?
+module load R/4.2.1-foss-2022a
```
sof202, line 66: I've had a look at the associated R script. It should still work for R version 4.2.1. Cool

sequencing/ATACSeq/jobSubmission/2_batchCalcQCMetrics.sh
```diff
@@ -6,46 +6,68 @@
...
 
-## move log files into a folder
-cd ${SCRIPTDIR}/ATACSeq/logFiles/${USER}
-mkdir -p ${SLURM_ARRAY_JOB_ID}
-mv *${SLURM_ARRAY_JOB_ID}_${SLURM_ARRAY_TASK_ID}.* ${SLURM_ARRAY_JOB_ID}
+echo "Running step 2 of ATAC-seq pipeline: Post alignment processing (QC metrics and fragment distribution)."
+echo "Calculating QC metrics and fragment distribution for samples in batch " ${SLURM_ARRAY_TASK_ID} 
+echo "Output directory is " "${ALIGNEDDIR}/QCOutput/"
```
sof202, line 70: This would prove rather useful when debugging. Thank you for this addition

sequencing/ATACSeq/jobSubmission/2_batchCalcQCMetrics.sh
```diff
@@ -6,46 +6,68 @@
 #SBATCH --nodes=1 # specify number of nodes.
 #SBATCH --ntasks-per-node=16 # specify number of processors per node
 #SBATCH --mail-type=END # send email at job completion 
-#SBATCH --output=ATACSeq/logFiles/%u/calcATACQC-%A_%a.o
-#SBATCH --error=ATACSeq/logFiles/%u/calcATACQC-%A_%a.e
-#SBATCH --job-name=calcATACQC
+#SBATCH --output=ATACSeq/logFiles/%u/ATACcalcQCS2-%A_%a.log
+#SBATCH --error=ATACSeq/logFiles/%u/ATACcalcQCS2-%A_%a.err
```
sof202, line 10: Could you remove %u from these so that it doesn't require the user to have created their own log files directory? The movement of log files later in the script accounts for the files not being in the correct spot yet. Considering the log files are going to be moved at the start of run time, these log files can be placed anywhere.

If the path were to be removed entirely:
```
#SBATCH --output=ATACcalcQCS2-%A_%a.log
#SBATCH --output=ATACcalcQCS2-%A_%a.err
```

Then, the script can be ran from anywhere (and won't fail if you don't run from the root of the repository)

[Sorry if this is a repeat comment]

sof202 (changes requested):
Great stuff. Could you update the name of the pull request (currently just says that step 1 is being updated).

sequencing/ATACSeq/jobSubmission/2_batchCalcQCMetrics.sh
marinafloresp: I also need to update the associated scripts

sequencing/ATACSeq/jobSubmission/2_batchCalcQCMetrics.sh
```diff
@@ -6,8 +6,8 @@
 #SBATCH --nodes=1 # specify number of nodes.
 #SBATCH --ntasks-per-node=16 # specify number of processors per node
 #SBATCH --mail-type=END # send email at job completion 
-#SBATCH --output=ATACSeq/logFiles/%u/ATACcalcQCS2-%A_%a.log
-#SBATCH --error=ATACSeq/logFiles/%u/ATACcalcQCS2-%A_%a.err
+#SBATCH --output=ATACcalcQCS2-%A_%a.log
+#SBATCH --error=ATACcalcQCS2-%A_%a.err
```
sof202, line 10: On line 52, the mv command needs to change for this to work. Replace 
`./ATACSeq/logFiles/${USER}/`
with
`${SLURM_SUBMIT_DIR}`

sequencing/ATACSeq/preprocessing/alignment.sh
```diff
@@ -1,43 +1,48 @@
...
-## OUTPUT
-# *_statsperchr.txt
-# *_sorted.bam
-# *_sorted_chr1.bam
-# *_dupMetrics.txt
-# *_depDup_q30.bam
-# *_postFilter_statsperchr.txt
-
-## (command line) VARIABLES
+## ===================================================================================================================##
```
sof202, line 1: I know this script is being ran using another one. But just in case the script is ran on its own in the future, could you put a shebang line for this script as well?

sequencing/ATACSeq/preprocessing/alignment.sh
```diff
@@ -1,43 +1,48 @@
...
-echo ${f[0]}
-echo ${f[1]}
+echo "Found trimmed files:" ${f[0]} ${f[1]}
+
+## =============== ##
+##    ALIGNMENT    ##
+## =============== ##
 
-## checks if last file exists and is not empty
+## checks if aligned file for input sample already exists
```
sof202, line 45: Why does it do this? Please add this into the comment

sequencing/ATACSeq/preprocessing/alignment.sh
```diff
@@ -64,9 +69,9 @@ then
 	# remove reads with q < 30, unmapped, mate unmapped, secondary alignment, reads failing platform
 	# only keep properly paired reads
 	echo "filtering aligned reads"
-    samtools view -F 524 -f 2  -q 30 -u ${ALIGNEDDIR}/${sampleName}_noMT.bam | samtools sort -n /dev/stdin -o ${ALIGNEDDIR}/${sampleName}_q30.tmp.nmsrt.bam
-    samtools view -h ${ALIGNEDDIR}/${sampleName}_q30.tmp.nmsrt.bam | $(which assign_multimappers.py) -k $multimap --paired-end | samtools fixmate -r /dev/stdin ${ALIGNEDDIR}/${sampleName}_q30.tmp.nmsrt.fixmate.bam
-	
+  samtools view -F 524 -f 2  -q 30 -u ${ALIGNEDDIR}/${sampleName}_noMT.bam | samtools sort -n /dev/stdin -o ${ALIGNEDDIR}/${sampleName}_q30.tmp.nmsrt.bam
+  samtools view -h ${ALIGNEDDIR}/${sampleName}_q30.tmp.nmsrt.bam | $(which assign_multimappers.py) -k $multimap --paired-end | samtools fixmate -r /dev/stdin ${ALIGNEDDIR}/${sampleName}_q30.tmp.nmsrt.fixmate.bam
```
sof202, line 73: I never liked this line in the original script. Does this actually work? It requires `assign_multimappers.py` to be on $PATH. For me at least, this is not the case. I feel like it would be better to just put the path to this python file into the config file:
```
MULTIMAPPER_PY_PATH="path/to/assign_multimappers.py"
```

sequencing/ATACSeq/preprocessing/alignment.sh
```diff
@@ -88,12 +93,7 @@ then
    samtools index ${ALIGNEDDIR}/${sampleName}.filt.nodup.bam
    
    ## get sum stats post to filtering
-   samtools stats ${ALIGNEDDIR}/${sampleName}.filt.nodup.bam	> ${ALIGNEDDIR}/QCOutput/${sampleName}filt.nodup.stats
+   samtools stats ${ALIGNEDDIR}/${sampleName}.filt.nodup.bam	> ${ALIGNEDDIR}/QCOutput/${sampleName}.filt.nodup.stats
```
sof202, line 96: nice spot

sequencing/ATACSeq/jobSubmission/0_setUp.sh
```diff
@@ -0,0 +1,174 @@
...
+fi
+
+## DIRECTORIES ##
+
+## Check that the metadata directory exists and contains sample list.
+if [[ ! -d ${METADIR} ]]
+then
+	echo 'METADATA directory does not exist'
+	exit 1
+elif  test -f ${METADIR}/samples.txt; then
```
sof202, line 77: Could you change this line to:
`elif [[ -f ${METADIR}/samples.txt ]]; then

So that the conditionals in this script are consistent?

sequencing/ATACSeq/jobSubmission/0_setUp.sh
```diff
@@ -0,0 +1,174 @@
...
+##    CHECKS    ##
+## ============ ##
+
+## CONFIG FILE ##
+
+## Check if config file for input project exist
+
+if [[ -s "/lustre/projects/Research_Project-MRC190311/ATACSeq/$1/config.txt" ]]
+then
+  source "/lustre/projects/Research_Project-MRC190311/ATACSeq/$1/config.txt"
```
sof202, line 55: Need another opinion on this one. But I'd say instead of forcing the config file to be in the root of the project directory. You could use the find command to check for the existence of the config file. Then you could use the output of the find command to source the file. This would make the placement of the config file less strict (some users might want to put the config files in a separate directory).

sequencing/ATACSeq/jobSubmission/0_setUp.sh
```diff
@@ -0,0 +1,174 @@
...
+	exit 1
+elif  test -f ${METADIR}/samples.txt; then
+	mapfile -t SAMPLEIDS < ${METADIR}/samples.txt 
+	echo "Sample list found. Number of sample IDs found:" """${#SAMPLEIDS[@]}"""
+else
+	echo 'samples.txt file does not exist in METADIR'
+fi
+
+## Check that the sampleSheet exist
+if [[ -d ${METADIR} ]] && [[ -s "${METADIR}/sampleSheet.csv" ]]
```
sof202, line 85: This first test of -d ${METADIR} has already been tested for on line 73. So isn't doing anything

sequencing/ATACSeq/jobSubmission/0_setUp.sh
```diff
@@ -0,0 +1,174 @@
...
+	then 
+		echo ${type[x]} "directory:" ${dirs[x]}
+	elif [ ${dirs[x]} = $RAWDATADIR ]; then 
+		echo ${dirs[x]} "directory does not exist"
+		exit 1
+	else
+		echo "Creating directory: " ${dirs[x]}
+		mkdir -p ${dirs[x]} 
+	fi
+done
```
sof202, line 109: Great use of an array here to make the directory checking more precise. Nicely done

sequencing/ATACSeq/jobSubmission/0_setUp.sh
```diff
@@ -0,0 +1,174 @@
...
+## MODULES ##
+
+## Check that the SAMstats environment exists and create it if not
+
+module load Anaconda3
+if [ -d ${ENVDIR} ]
+then 
+	echo "ENCODEQC environment exists"
+  source $CONDAENV
+  conda activate ${ENVDIR}
```
sof202, line 120: This confuses me. Are you trying to activate the directory where the conda environments are held? If so, this command doesn't work. You'll get:
```
not a conda environment
```

If it is not a directory, then the -d in the test above won't work. Could you explain what is happening here?

sequencing/ATACSeq/jobSubmission/0_setUp.sh
```diff
@@ -0,0 +1,174 @@
...
+## Check that pip environment exist and create it if not
+module load Python/3.9.6-GCCcore-11.2.0-bare
+
+if [[ -d ${PIPENV} ]]
+then
+  echo "Pip environment exists."
+  source /lustre/home/${USER}/pythonEnv/bin/activate
+  echo "List of packages included in pip environment is output in project directory"
+  pip list > "/lustre/projects/Research_Project-MRC190311/ATACSeq/$1/pipEnv.txt"
+  deactivate
```
sof202, line 145: This also confuses me. the pip environment exists, but you are specifically sourcing 'pythonEnv' and not ${PIPENV}. ${PIPENV} isn't used at all in this part of the if block, why is it in the test?

sequencing/ATACSeq/jobSubmission/0_setUp.sh
marinafloresp: You are right, need to specify the name of the conda env not the directory

sequencing/ATACSeq/preprocessing/setUpRenv.r
```diff
@@ -0,0 +1,32 @@
...
+##                                                                                             ||
+## INPUTS:                                                                                     ||
+## - <project> : project on which analysis is being run                                        ||
+##                                                                                             ||
+## REQUIRES:                                                                                   ||
+## - R/4.2.1-foss-2022a                                                                        ||
+## ============================================================================================##
+
+args <- commandArgs()
+user <-args[6]
```
sof202, line 17: Why args[6]?

sequencing/ATACSeq/preprocessing/setUpRenv.r
```diff
@@ -0,0 +1,32 @@
...
+packagesBase <- c("diptest", "ptest", "plyr","dplyr","ggplot2","ggpubr","RColorBrewer",  "pheatmap", "gridExtra", "FME", "data.table", "readr","scales","tibble","kableExtra","tidyverse","rstatix","conflicted","knitr","vioplot",
+"corrplot","reshape2","cowplot")
+packagesBiocM <- c("BiocManager","ATACseqQC","Rsubread","GenomicRanges","ChIPpeakAnno","ChIPseeker","TxDb.Hsapiens.UCSC.hg38.knownGene","org.Hs.eg.db","edgeR","DiffBind","csaw","GenomicAlignments", "GenomicTools.fileHandler", 
+"rtracklayer", "clusterProfiler","ReactomePA")
+
+## As most packages are installed through BiocManager, we first installed this if not found
+if (!require("BiocManager", quietly = TRUE))
+    install.packages("BiocManager")
+BiocManager::install(version = "3.16")
+BiocManager::install(packagesBiocM, lib = paste0("/lustre/home/",user,"/R/x86_64-pc-linux-gnu-library/4.2") )
```
sof202, line 29: Will this make it past the proxy server? Just because of recent compilcations

sequencing/ATACSeq/preprocessing/setUpRenv.r
```diff
@@ -0,0 +1,32 @@
...
+
+## Then we install the rest of packages if not installed already
+install.packages(setdiff(packagesBase, rownames(installed.packages())),lib = paste0("/lustre/home/",user,"/R/x86_64-pc-linux-gnu-library/4.2"))
```
sof202, line 32: I like this way of doing things. Much more succinct than the way I'm currently handling installing packages. I hope to use a similar approach myself

sof202 (changes requested):
There's a couple of changes (or at least clarifications) that I'd like to see first, but this is a great improvement to the ATAC pipeline

sequencing/ATACSeq/preprocessing/setUpRenv.r
marinafloresp: I would need to double check this installation, as I already have that installed and reinstalling is avoided if not forced.

sof202:
Also, this pull request is getting quite large now (over 400 lines changed). As I'm following along with this PR since it was created it's not too bad. But in future, I would consider splitting the PR up (into multiple separate PRs). This is just to increase your chances of someone completing a code review (as it is less likely someone will spend 4 hours on a code review than 30 minutes)

Issue #87, opened 2024-04-04
[Documentation Fault]: Remove delimiters<> [] ()


sof202:
### Fault type

Rewording required

### Please describe the fault in detail

As other developers have worked through the 'How to contribute guide', it came to my attention that code blocks of the form:

```bash
git branch <your-branch>
```

were interpreting the <>s literally. As such, these need to be removed to improve understandability.

PR #92, opened 2024-04-04
Remove delimiters across developer pages

sof202:
# Description

This pull request will remove delimiters like [], <> (*etc.*) from developer pages. I apologise for any confusion that resulted from these, this is a habit I picked up from forums I used as a child.

## Issue ticket number

This pull request is to address issue: #87.

## Type of pull request

- [ ] Bug fix
- [ ] New feature/enhancement
- [ ] Code refactor
- [x] Documentation update

## Checklist

- [x] I have made the corresponding changes to the documentation


marinafloresp:
What about using italics instead?

sof202:
> What about using _italics_ instead?

I like the idea but:
You can't put ** inside quote blocks:
```bash
*code*
```

However, you *could* use <pre> tags. But these come with their own problems. Namely: no horizontal scrollbar (for longer code blocks) and no syntax highlighting support (for docusaurus).

ejh243 (approved):

Issue #137, opened 2024-05-14
[Bug]: add scripts directory variable to config


ejh243:
### Contact Details

e.j.hannon@exeter.ac.uk

### What happened?

When submitting the DNAm QC job submission script, fails due to incorrect paths.

We will fix by adding a variable for the location of the scripts folder to the config file and remove the need for relative paths that often fail.

Issue #52, opened 2024-02-26
[Enhancement suggestion]: add script to ensure user has all R packages pre installed


ew367:
### Enhancement suggestion

add script at the beginning that checks if user has the necessary R packages installed, and in the correct versions. If not, install them prior to running the QC pipeline

ejh243:
@rhaigh5 has started this here: https://github.com/ejh243/BrainFANS/blob/master/array/DNAm/preprocessing/installLibraries.r

Is there anything missing?

Issue #43, opened 2024-02-21
[Bug]: CRLF used in array\SNPArray\PennCNVEURonly.sh


sof202:
### What happened?

Windows EOL is used in the file:
array/SNPArray/PennCNVEURonly.sh

All other files in the repository use \n (LF) EOL. But this file uses \r\n (CRLF) EOL. 

When ran on any linux system (which the pipeline is designed for), the character ^M will be placed onto the end of every line that is read by the bash interpreter (which can lead to unwanted behaviour).

### How can the bug be reproduced?

Run this file on any linux machine, you will see errors for most lines.

Ignore the error commands associated with the inability to open files (this is due to no configuration file being inputted for this run).

### Relevant log output

```shell
: No such file or directorycd: /
: No such file or directorycd: SNPdata/Merged
PennCNVEURonly.sh: line 8: $'\r': command not found
grep: PredictedPopulations.csv: No such file or directory
tail: cannot open 'EURSamplesTmp.txt' for reading: No such file or directory
PennCNVEURonly.sh: line 12: $'\r': command not found
: No such file or directory cd: /SNPdata/CNV
PennCNVEURonly.sh: line 15: $'\r': command not found
PennCNVEURonly.sh: line 17: /compile_pfb.pl: No such file or directory
PennCNVEURonly.sh: line 18: $'\r': command not found
: No such file or directory cd: /SNPdata/CNV
: No such file or directory cd: PennCNVOutput
: No such file or directory cd: ..
PennCNVEURonly.sh: line 24: $'\r': command not found
: No such file or directory cd: PennCNVInput/
PennCNVEURonly.sh: line 27: /detect_cnv.pl: No such file or directory
PennCNVEURonly.sh: line 27: $'\r': command not found
PennCNVEURonly.sh: line 28: $'\r': command not found
PennCNVEURonly.sh: line 30: /filter_cnv.pl: No such file or directory
PennCNVEURonly.sh: line 31: $'\r': command not found
PennCNVEURonly.sh: line 32: $'\r': command not found
PennCNVEURonly.sh: line 33: $'\r': command not found
PennCNVEURonly.sh: line 34: $'\r': command not found
PennCNVEURonly.sh: line 35: $'\r': command not found
```

sof202:
The same issue exists for:

sequencing\DNAhydroxy\jobSubmission\peakCalling5hmCIncreasingSampleSizes.sh

PR #139, opened 2024-05-14
137 bug add scripts directory variable to config

### LINKED ISSUE #117: [Documentation Fault]: DNAm QC add config details
### Contact Details

e.j.hannon@exeter.ac.uk

### Fault type

Incorrect/outdated information given

### Please describe the fault in detail

Need to add details on where the config files is stored in [array/DNAm/preprocessing/loadDataGDS.r](https://github.com/ejh243/BrainFANS/blob/master/array/DNAm/preprocessing/loadDataGDS.r) and likely others.

ejh243:
# Description

This pull request will simplify use of config files with DNAm pre-processing pipeline and remove need to use relative paths when executing scripts. 

## Issue ticket number

This pull request is to address issue: #137 .

Also closes #117, #52, #145

## Type of pull request

- [X] Bug fix
- [ ] New feature/enhancement
- [X] Code refactor
- [X] Documentation update

## Checklist


AnthonyKl98 (approved):
These changes simplify the process of running the QC since only the file path to the config file is required on the command line. I'm able to run the QC in this branch, resolving previous issue related to changing directories

array/DNAm/config/config.example.txt
```diff
@@ -1,10 +1,13 @@
 ## Data filepaths
-export DATADIR=<data path>${PROJECT}
+export DATADIR=<data path>
```
sof202, line 2: In recent past I've found that people get confused with <example here> as some think that the "<" and ">" characters should be kept. I see a lot of documentation use this instead:

```
path/to/data
```

which is more intuitive I guess.

array/DNAm/preprocessing/checkColnamesFunction.r
```diff
@@ -0,0 +1,51 @@
...
+		}
+		metrics[['presentIDs']] <- pres
+		
+		#b) find absent cols
+		ab <- exp_cols[exp_cols %ni% colnames(sampleSheet)]
+		if(verbose){print(c("Absent colnames:",ab))}
+		metrics[['absentIDs']] <- ab
+		
+		#c) find close matches
+		near <- exp_cols[amatch(colnames(sampleSheet), exp_cols, maxDist = 2)] # maxDist = number of characters difference
```
sof202, line 29: Is this `amatch` from the stringdist package? My version of R doesn't recognise this function and I don't see any namespaces being loaded.

sof202: Also, it might be worth stripping `exp_cols` and `colnames(sampleSheet)` of any common separators (".", "_", "-" *etc.*) and making them all lower case first. Then you can be more stringent on `maxDist`. 

Currently the matching algorithm is set to "Damerau–Levenshtein distance" I believe. If you have any similarly named columns in any sampleSheets (never worked with one so this might not be an issue), then this might falsely match two columns.

array/DNAm/preprocessing/checkColnamesFunction.r
```diff
@@ -0,0 +1,51 @@
...
+			if(nrow(nrmatch)>0){
+				if(verbose){print(paste("Identified", nrow(nrmatch), "near match(es)..."))}
+				if(verbose){print(nrmatch)}
+				metrics[['nearMatchIDs']] <- nrmatch
+			}
+		}
+		
+	}
+	if(!verbose){
+		return(metrics)
```
sof202, line 49: Why is this only returning metrics when verbose is set to `FALSE`?

sof202: Is it just so that you can call the function in 'checkColnamesSampleSheet.r' without `metrics` being sent to the standard output when you only want the messages printed? If so, you could always call the function with the `invisible()` wrapper:

```R
invisible(checkColnames(sampleSheet, req_cols, type='Required'))
```

`invisible` will mask any return values, but won't mask print messages that are inside a function.

array/DNAm/preprocessing/checkColnamesFunction.r
alicemfr: Re your first point, I see what you mean. But since it is important the column names are written with the correct separator (underscore), a column called Cell.Type might incorrectly match to Cell_Type because we've stripped the separator. What do you think?

Re your second point, hopefully this answers it but let me know if not. 
Yes the sample sheet could have 2 column names that are similarly named for example Cell_type and Cell_Type. 
If "Cell_Type" is the required column name then this will match perfectly to one of the columns and will be flagged in `pres`. When it comes to the near matches, amatch picks up both near matches and exact matches so both Cell_Type and Cell_type will be flagged again here. 
Since in this section we are interested in near matches I figured that we would want to remove from the printed output any cases of exact matches (see variable `equal`). 
And if this happens to be a case where there is an exact match AND a near match (such as Cell_type and Cell_Type), I don't think we need to flag anything wrong here because it has met the criteria of having the required column name (Cell_Type). Therefore I find and remove them both (see variable `other`). This is what I attempt to do in lines 34-39.

sof202: Oh ok, I think my understanding of the original problem was wrong.

I though you were trying to account for column names being inconsistent between sample sheets. So for example, in one sample sheet you might have `Cell_Type` whilst in another you might have `cell.type` (where these are actually the same column, displaying the same kind of information, just with different names). 

That's why I made those suggestions. As `Cell_Type` and `cell.type` have a distance of 3 from one another. If you remove separators and make them lower case they are the same (so you can be more strict on the maxDist). 

My bad.

alicemfr: Ah I see!

array/DNAm/preprocessing/checkColnamesFunction.r
alicemfr: Thanks for suggesting this, I'll look into it as this might be what I need.

array/DNAm/preprocessing/jobSubmission/1_runDNAmQC.sh
```diff
@@ -65,4 +66,13 @@ Rscript CETYGOdeconvolution.r ${DATADIR}
...
-date -u
\ No newline at end of file
+date -u
+
+mkdir -p ${DATADIR}/logFiles
+
+mv "${SLURM_SUBMIT_DIR}/QCDNAdata_${SLURM_JOB_ID}.log" \
+"${DATADIR}/logFiles/QCDNAdata_${SLURM_JOB_ID}.log"
+mv "${SLURM_SUBMIT_DIR}/QCDNAdata_${SLURM_JOB_ID}.err" \
+"${DATADIR}/logFiles/QCDNAdata_${SLURM_JOB_ID}.err"
```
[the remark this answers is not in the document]
sof202: There's not really a good way around this though. Maybe worth putting into READMEs if this ever becomes relevant.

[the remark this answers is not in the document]
ejh243: Yes I did think this. That's partly why I include the print statements so the user knows where to look for the log files. I didn't have any obvious ideas how to avoid this?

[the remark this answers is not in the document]
sof202: I mean it's not a problem at all for this script as the only way it will exit early is if the user cancels the job.

It usually becomes a problem when you have fail safe exit statements like:

```bash
cd some/directory/change || exit 1
```

Also, the print statements are kinda useless no? They will only ever be in the log files themselves. So if you are reading the line that tells you where the log file is, you've already found it.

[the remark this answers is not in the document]
ejh243: oh yeah!

array/DNAm/preprocessing/installLibraries.r
```diff
@@ -42,41 +38,34 @@ if(arrayType=='450K'){
...
 
 #---------------------------------------------------------------------#
 # INSTALL PACKAGES FOR QC METRICS
 #---------------------------------------------------------------------#
 
-install.packages("e1071")
-install.packages("stringdist")
-install.packages("data.table")
-
+install.packages(setdiff(c("e1071","stringdist","data.table"), rownames(installed.packages())),repos = "https://cran.r-project.org")
```
sof202, line 55: This is really clean, thank you @marinafloresp for the suggestion

sof202:
Just to let you know @alicemfr, when you are saving `array/DNAm/preprocessing/installLibraries.r`, you are saving it with carriage return sequences (CRLF end of line). Make sure your text editor is using LF as otherwise the files you are changing might break.

image

(^M is a literal carriage return, #43)

sof202:
The same goes for:
array/DNAm/preprocessing/checkColnamesFunction.r
array/DNAm/preprocessing/README.md  (though this isn't being executed so it doesn't matter)
array/DNAm/preprocessing/checkColnamesSampleSheet.r 
array/DNAm/preprocessing/normalisation.r

array/DNAm/preprocessing/checkColnamesSampleSheet.r
ejh243, line 1: Do we want this file to a) just report to the user the colnames are incorrect and stop execution or b) fix the colnames and continue running?

array/DNAm/preprocessing/installLibraries.r
```diff
@@ -1,81 +1,104 @@
...
-install_github("EpigeneticsExeter/cdegUtilities")
+
+##INSTALL LIBRARIES SCRIPT
+##Installs all R packages required DNAm QC pipeline
+
+##---------------------------------------------------------------------#
+##
+## Title: Install required packages
+##
+## Purpose of script: installs all packages required for QC pipeline steps.
```
ejh243, line 9: specifically the DNAm QC

array/DNAm/preprocessing/checkColnamesSampleSheet.r
alicemfr: My thought is that it would be risky to assume we know what the user's column name means. Even if it comes up as a close match I think it should be up to the user to make that call. So my preference would be (a).

To make that happen, we could force the script to stop running, with an error message like: "Not all of your column names match the required format. Please edit sample sheet, noting any near matches, and rerun.". 
But in the submission script even if checkColnamesSampleSheet.r stops it will still move on to loadDataGDS.r. Can we stop that from happening with `exit` or something?

ejh243:
I have tested and am happy so am going to merge.

Issue #148, opened 2024-05-30
[Bug]: FASTP detecting adapters but not trimming


marinafloresp:
### Contact Details

m.flores-payan@exeter.ac.uk

### What happened?

The fastp.sh script in the preScripts folder uses FASTP to perform trimming on raw reads and produce fastqc on the results. However, even if the --detect_adapter_for_pe parameter is set, not all reads with Nextera adapter are detected and they failed to be trimmed. Nextera adapters should be detected and properly trimmed from reads. 

### How can the bug be reproduced?

Run the fastp.sh file using only the --detect_adapter_for_pe. If fastqc is used on the trimmed output files, adapters are still detected.

Issue #147, opened 2024-05-29
[Enhancement suggestion]: Add QC step after trimming and before alignment


marinafloresp:
### Contact Details

m.flores-payan@exeter.ac.uk

### Enhancement suggestion

Currently, there is not a qc step after trimming of reads is done in the ATACseq pipeline. This enhancement will allow the user to check trimming statistics and make sure the process has run as expected. In addition, when multiqc is done in step 4, trimming qc results should be collated in a single report as done with raw reads qc and alignment qc.

PR #149, opened 2024-05-31
Config file examples for pipeline

marinafloresp:
# Description

This pull request will update the example config file required to run the ATACseq pipeline. Scripts used in step 0 and 1 of the pipeline have been update accordingly.

## Issue ticket number

This pull request is to address issue: #148 and #147 .

## Type of pull request

- [X] Bug fix
- [X] New feature/enhancement
- [ ] Code refactor
- [ ] Documentation update

## Checklist


sequencing/ATACSeq/preprocessing/fastp.sh
```diff
@@ -0,0 +1,61 @@
...
+
+##Specific adapters for R1 and R2 need to be specified for correct trimming
+adapterR1=CTGTCTCTTATACACATCTCCGAGCCCACGAGAC
+adapterR2=CTGTCTCTTATACACATCTGACGCTGCCGACGA 
+
+## ============ ##
+##    TRIM      ##
+## ============ ##
+
+fastp --detect_adapter_for_pe --adapter_sequence=$adapterR1 --adapter_sequence_r2=$adapterR2 --length_required=27 --thread=8 --in1=${f1} --in2=${f2} --out1=${TRIMDIR}/${outf1} --out2=${TRIMDIR}/${outf2} --html=${TRIMDIR}/qc/${sampleName}_fastp.html --json=${TRIMDIR}/qc/${sampleName}_fastp.json
```
sof202, line 55: Why are you specifying the worker thread number here (--thread=8)? Considering you are parallelising the trimming process, you could base the worker thread off of the array ID for a speed up.

So you could use modular arithmetic here right? Array task 1 goes to thread 1, task 2 to thread 2, ..., task 16 to thread 16, task 17 back to thread 1 etc.

This way you can take advantage of running the job in parallel, instead of running everything on the same thread.

sof202 (changes requested):

sequencing/ATACSeq/preprocessing/fastp.sh
```diff
@@ -0,0 +1,61 @@
...
+
+## create output filenames
+outf1=${f1/.f/_trimmed.f}
+outf2=${f2/.f/_trimmed.f}
+
+mkdir -p ${TRIMDIR}/qc/
+
+##Specific adapters for R1 and R2 need to be specified for correct trimming
+adapterR1=CTGTCTCTTATACACATCTCCGAGCCCACGAGAC
+adapterR2=CTGTCTCTTATACACATCTGACGCTGCCGACGA
```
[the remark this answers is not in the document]
sof202: I don't know enough about ATAC, maybe this is set in stone for all datasets

sequencing/ATACSeq/preprocessing/fastp.sh
sof202:
```bash
--thread=$(( (${SLURM_ARRAY_TASK_ID} % 16) + 1 ))
```

The `+ 1` here is so that you don't get `--thread=0` but instead `--thread=1-16`. I think you'd need to parse in the $SLURM_ARRAY_TASK_ID into this script as I'm not sure what the scope of this variable is.

sequencing/ATACSeq/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+##### -scripts executed-
+- ATACSeq/preprocessing/intallLibraries.sh : checks for already installed R libraries and install if not found.
+
+##### -parameters-
+- `--array`: should be 0 for this script as general analysis is run rather than individual samples.
+- `(project-name)` project's directory.
+
+##### -requires-
+- packagesPip.txt, environment.yml and config.r file paths to be specified in the config.txt file.
```
ejh243, line 34: Do these need to be specified in the config file, as they are in this repo? If the scripts folder is in the config file, you can build the path from these elements.

sequencing/ATACSeq/README.md
```diff
@@ -0,0 +1,201 @@
...
+- `(project-directory)` project's directory.
+-optional-
+- `[STEPS]` They may be combined, with desired steps included as single string, i.e. MULIQC,COLLATE. Default if left blank is to run all of them.
+  - `SPLIT`: Generate only samples with X and Y chromosomes reads.
+  - `PEAKS`: Perform only peak calling on new sex chromosomes reads samples.
+  - `CHECK`: Collate peak calling on sex chromosomes results and check assigned sex of each sample.
+
+### 6. Genotype check
+
+  `sbatch --array=<number of batch jobs> ATACSeq/jobSubmission/6_batchRunGenotypeConcordance.sh (project directory) [STEPS]`
```
ejh243, line 142: do we need to provide guidance on how many batch jobs need to be specified? If I recall it is the same as the number of samples?

sequencing/ATACSeq/config/config.txt
```diff
@@ -1,14 +1,39 @@
...
+##                                                            ||
+## The project's folder name needs to be specified in the     ||
+## first variable. Please specify the modules versions to be  ||
+## used.                                                      ||
+##                                                            ||
+## ===========================================================##
+
 ## Sequencing data filepaths
+export PROJECT="project name"
 export DATADIR=/lustre/projects/Research_Project-MRC190311/ATACSeq/${PROJECT}
```
ejh243, line 18: Can we change these to be more generic (i.e remove reference to ISCA file paths)?

sequencing/ATACSeq/config/config.txt
```diff
@@ -1,14 +1,39 @@
...
+export CONFIGR=/lustre/projects/Research_Project-MRC190311/ATACSeq/${PROJECT}/config.r
```
ejh243, line 19: As we expect this to be in the DATADIR specified above we can build it from that variable.

sequencing/ATACSeq/config/config.txt
```diff
@@ -23,16 +48,15 @@ export XCHRBED=/lustre/projects/Research_Project-MRC190311/references/GeneLists/
 export UCSCRef=/lustre/projects/Research_Project-MRC190311/references/UCSC/
 export GENOMESIZE=/lustre/projects/Research_Project-MRC190311/references/grch38/genome.info
 export SEGREFGENOME=/lustre/projects/Research_Project-MRC190311/references/Segway/GRCh38/Sequence/WholeGenomeSegwayFasta/genome.char.fa
-
+export SNP=${GENODIR}/hg38/allchr_filt_rsq_maf.vcf.gz
```
ejh243, line 51: somewhere (README I think) we need to clarify what this file is, how it is created etc. Also how robust is the pipeline if it is not present?

sequencing/ATACSeq/jobSubmission/2_batchCalcQCMetrics.sh
```diff
@@ -6,44 +6,64 @@
 #SBATCH --nodes=1 # specify number of nodes.
 #SBATCH --ntasks-per-node=16 # specify number of processors per node
 #SBATCH --mail-type=END # send email at job completion 
-#SBATCH --output=ATACSeq/logFiles/%u/calcATACQC-%A_%a.o
-#SBATCH --error=ATACSeq/logFiles/%u/calcATACQC-%A_%a.e
-#SBATCH --job-name=calcATACQC
+#SBATCH --output=ATACcalcQCS2-%A_%a.log
```
ejh243, line 9: Can you specify in the README, where the log files are located.

sequencing/ATACSeq/preprocessing/fastp.sh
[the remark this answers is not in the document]
ejh243: Hmm I think probably better in config file in case these do need to be changed?

sequencing/ATACSeq/preprocessing/intallLibraries.r
```diff
@@ -0,0 +1,33 @@
+## ============================================================================================##
+##                     ATAC-seq pipeline STEP 0.2: Set up R environment                        ##
+## ============================================================================================##
+## EXECUTION: Rscript ./sequencing/ATACSeq/preprocessing/setUpRenv.sh  <user>                  ||
+## - execute from scripts directory                                                            ||
+##                                                                                             ||
+## DESCRIPTION: This script checks if required R packages are installed and install them if not//
+##                                                                                             ||
+## INPUTS:                                                                                     ||
+## - <project> : project on which analysis is being run                                        ||
```
ejh243, line 10: isn't this user?

sequencing/ATACSeq/preprocessing/intallLibraries.r
```diff
@@ -0,0 +1,33 @@
...
+## EXECUTION: Rscript ./sequencing/ATACSeq/preprocessing/setUpRenv.sh  <user>                  ||
+## - execute from scripts directory                                                            ||
+##                                                                                             ||
+## DESCRIPTION: This script checks if required R packages are installed and install them if not//
+##                                                                                             ||
+## INPUTS:                                                                                     ||
+## - <project> : project on which analysis is being run                                        ||
+##                                                                                             ||
+## REQUIRES:                                                                                   ||
+## - R/4.2.1-foss-2022a                                                                        ||
```
ejh243, line 13: Can we frame as R version > XXX?

sequencing/ATACSeq/preprocessing/sampleHMMRatac.sh
ejh243, line 1: Given this regularly returns 0 peaks at a sample level, shall we exclude it to speed up the process?

ejh243 (commented):
I haven't gone through everything in great detail but just looked for a few keys bits.

sequencing/ATACSeq/jobSubmission/2_batchCalcQCMetrics.sh
sof202: I'm pretty sure these log files work the same way now as the DNAm pipeline (though they are for an array here instead of a single job). You can likely just adapt what Eilis has already written for that pipeline here (this will make the repository more consistent).

Eilis' explanation below:

>During execution log files are temporaily located in the folder you submitted the script from with the filenames QCDNAdata_XXX.log and QCDNAdata_XXX.err, where XXX is replaced with the job number provided by your HPC scheduler. Please ensure you have the correct permissions to create files in this folder, otherwise the job will instantly fail. If the job fails before the script is run to completion, this is where you need to look for troubleshooting. Once the QC pipeline is complete these files are moved to your data directory in a folder called `logFiles`.

sof202:
I can't comment it directly, but Line 4 in `intallLibraries.sh` is I presume outdated.

> .## EXECUTION: Rscript ./sequencing/ATACSeq/preprocessing/setUpRenv.sh  <user>                  ||

This script (setUpRenv.sh) doesn't appear to exist anymore. I'm assuming this is instead supposed to be:

> .## EXECUTION: ./sequencing/ATACSeq/jobSubmission/0_setup.sh  <user>                  ||

sequencing/ATACSeq/config/config.txt
```diff
@@ -13,65 +13,73 @@
...
-export DATADIR=/lustre/projects/Research_Project-MRC190311/ATACSeq/${PROJECT}
-export CONFIGR=/lustre/projects/Research_Project-MRC190311/ATACSeq/${PROJECT}/config.r
+export MAIN_DIR="/path/to/project/folder/${PROJECT}"
+export CONFIGR="/path/to/R/config/file/${PROJECT}/config.r"
 
-## Specify modules versions
+## Software/script filepaths
+export SCRIPTS_DIR="/path/to/this/scripts/folder"
+export SUB_SCRIPTS_DIR="${SCRIPTS_DIR}/path/to/subcripts/"
+export RSCRIPTS_DIR="${SCRIPTS_DIR}/Rscripts"
```
sof202, line 29: By scripts folder, I presume you mean this specific part of BrainFANS, not BrainFANS' root directory. I assume this because of line 29 here (using `$SCRIPTS_DIR/Rscripts`). If I am correct, could you make it so line 28 is just `$SCRIPTS_DIR/preprocessing` to make it easier on the user?

sof202: Also, I think I'm having the same problem in my config file templates. Do you think that it users might struggle to find which variables they actually have to change? 
Using the above as an example, it is hopefully obvious that `27` needs to be changed. But then you need to change `28` but not `29`. I think this might confuse some people considering the size of the config file. 

I haven't thought of a good solution for my own set of scripts yet. Maybe a key? I don't know. Just wanted to see if you share that sentiment.

marinafloresp: The only solution I can think of is to make the user changes as little variables as possible and at the same time expecting a minimum of consistency in folder names (like "Rscripts" or "0_metadata") or other parts of the pipeline. If we don't make that consistency to be expected by the pipeline, then the user will need to change all variables to accommodate to their own data structure. I see it as a never-ending circle.

sof202: Ah yeah, but what I mean was. If you only expect the user to only change a few lines in the configuration file (out of 30ish). How can we make it obvious to the user which ones they need to change? Obviously they should change the ones that say `path/to/scripts` or something similar in them. But these might not be seen when the whole file is a bunch of path variables. Maybe just putting something in the README as to which variables need to be changed by the user?

sequencing/ATACSeq/config/config.txt
```diff
@@ -1,41 +1,32 @@
...
+## ===========================================================================##
+
+## ========================= ##
+|| Project set up filepaths  ||
+## ========================= ##
+
+## Specify the name of the project's directory ##
 export PROJECT="project name"
+## Specify the full project's directory ##
 export MAIN_DIR="/path/to/project/folder/${PROJECT}"
```
sof202: Yeah this works well. Separating it out is probably the best solution

marinafloresp: I will add extra info in the README file to make very clear what needs to be changed and what datasets, references, etc are expected from the user

sequencing/ATACSeq/config/config.txt
```diff
@@ -44,42 +35,67 @@ export ALIGNED_DIR="${MAIN_DIR}/3_aligned"
...
+export REFGENOME="${REFERENCES_DIR}/reference_genome"
+export BLACKLIST="${REFERENCES_DIR}/blacklisted_peak_regions"
+export XCHRBED="${REFERENCES_DIR}/X_chr_genes"
+export PAR="${REFERENCES_DIR}/pseudoautosomal_regions"
+export GENOMEFASTA="${REFERENCES_DIR}/Whole_genome_fasta/"
+export KGREF="${REFERENCES_DIR}/reference_1000G/"
+export SNP="${REFERENCES_DIR}/SNP_files"
+export UCSCRef="${REFERENCES_DIR}/UCSC/"
+export GENOMESIZE="${REFERENCES_DIR}/genome.info"
+export SEGREFGENOME="${REFERENCES_DIR}/genome.char.fa/"
```
sof202, line 90: Is this a file or a directory?
Only asking because .fa is a file extension (FASTA)

marinafloresp: Yes, that is a file, the / is an error

sequencing/ATACSeq/config/config.txt
sof202: good idea

sof202:
Hi @marinafloresp, just noticed that you have named one of your scripts: `intallLibraries.r` instead of `installLibraries.r`. 

This typo might mean that this script is never being called?

```
sequencing/ATACSeq/Rscripts/intallLibraries.r
```

marinafloresp:
Is that in the README file? I am finishing up updating all the pipeline scripts, so I have probably already corrected that

sof202:
The file itself is named wrong, I'm not looking through the README or anything (might be a typo there as well, haven't checked). I just presume you meant to call the file `installLibraries.r`

PR #160, opened 2024-07-16
Update ATAC-seq pipeline with latest working scripts

marinafloresp:
# Description

This pull request will update the scripts used in the ATAC-seq pipeline within steps 0 to 6. Older unused scripts have been removed and scripts have been updated to the latest working version. The README file has been updated with extra information. The pipeline's main scripts now include more output messages to let the user know what processes have been successfully run and subscripts check whether the right output have been produced, and inform the user if not.

## Type of pull request

- [X] Bug fix
- [X] New feature/enhancement
- [X] Code refactor
- [X] Documentation update

## Checklist


sequencing/ATACSeq/jobSubmission/6_batchRunGenotypeCheck.sh
```diff
@@ -0,0 +1,157 @@
...
+## ============ ##
+##    SET UP    ##
+## ============ ##
+
+## print start date and time
+echo Job started on:
+date -u
+
+## load config file provided on command line related to the specified project
+source "${1}/config.txt"
```
sof202, line 53: In case the config file is not found (because the user typed it incorrectly or gave the wrong path *etc.*), could you put a error message here?

So you could do:
```bash
source "${1}/config.txt" || { echo "Could not find config.txt at $1"; exit 1; }
```

Or something to that effect.

marinafloresp: Yes, I will add that check

sequencing/ATACSeq/jobSubmission/6_batchRunGenotypeCheck.sh
```diff
@@ -0,0 +1,157 @@
...
+## ============ ##
+##    STEPS     ##
+## ============ ##
+
+## option COMPARE: genotype of samples are verified by comparing them with existing genotype information   
+if [ $# = 1 ] || [[ $2 =~ 'COMPARE' ]]
+then
+  module purge
+  module load $PICARDVERS
+  module load GATK
```
sof202, line 93: Is a specific version of GATK required? I don't really know what this is, but all other modules loaded have a specific version requirement in the config file. Why not this one?

sequencing/ATACSeq/jobSubmission/6_batchRunGenotypeCheck.sh
```diff
@@ -0,0 +1,157 @@
...
+  module load $SAMTVERS
+  
+  echo " "
+  echo "|| Running STEP 6.1 of ATAC-seq pipeline: COMPARE. Samples will be compared to their matching genotype data.||"
+  echo " "
+  echo "Output directory will be: ${ALIGNED_DIR}/genotypeConcordance"
+  echo "Output directory will be: ${ALIGNED_DIR}/baseRecalibrate"
+  echo " "
```
sof202, line 101: You might find the following useful for whenever you want to print a big chunk of text to the standard output:

```bash
cat << EOF
text
more text

text


|| text ||
EOF
```

Just put all of your text you want to print inside the two `EOF` keywords.
This saves you from putting all of these `echo` statements and `""` lines. Most Linux systems have `cat `installed so it should still be pretty portable. Look up `heredoc` for more information.

sequencing/ATACSeq/jobSubmission/6_batchRunGenotypeCheck.sh
```diff
@@ -0,0 +1,157 @@
...
+  
+  # process a line from IDMap file
+  IDS=($(head -n ${SLURM_ARRAY_TASK_ID} ${META_DIR}/matchedVCFIDs.txt | tail -1))
```
sof202, line 110: Just a quick check, can `${SLURM_ARRAY_TASK_ID}` here be 1? If it can be then the first row of `matchedVCFIDs.txt` is a header row and so this won't give you what you want. 

Looking at an example of `samples.txt` the file doesn't have a header row, so I presume this will be pulling out the incorrect sample for all task ids (provided that my assumption that  `${SLURM_ARRAY_TASK_ID}` corresponds to the sample in `samples.txt` with the same row number).

I might be being silly and this works regardless due to 0/1 indexing, just checking.

sequencing/ATACSeq/jobSubmission/6_batchRunGenotypeCheck.sh
```diff
@@ -0,0 +1,157 @@
...
+  module load $SAMTVERS
+  
+  echo " "
+  echo "|| Running STEP 6.3 of ATAC-seq pipeline: SWITCH. Samples with potential genotype contamination will be selected for potential switches.||"
+  echo " "
+  echo "Output directory will be: ${ALIGNED_DIR}/genotypeConcordance"
+  echo "Output directory will be: ${ALIGNED_DIR}/baseRecalibrate"
+  echo " "
+  
+  cd ${ALIGNED_DIR}/genotypeConcordance/
```
sof202, line 150: Does this directory necessarily exist? The only time I see the directory being created is in 
`compareBamWithGenotypes.sh` on line 94. However, that script is only being called if you enter `COMPARE` as the second argument to this script. If you don't have `COMPARE` but do have `SWITCH` in the second argument, this line will run (but the directory hasn't been created). 

I guess the directory might be created with line 42 in `searchBestGenoMatch.sh`. But this script won't have run yet (not until line 152 of this script). 

If this `cd` command fails, it looks like the only ramifications will be that lines 146 and 147 will likely fail (as they can't find these files).

sof202: Then again, if you were to create this directory here, then the `awk` commands below will still fail as these `*.selfSM` files won't be found regardless (as the directory would be empty).
Am I missing something here? It feels like these `awk` commands will fail if the `COMPARE` section is not ran.

sequencing/ATACSeq/subScripts/compareBamWithGenotypes.sh
```diff
@@ -1,34 +1,49 @@
...
+## ============ ##
+
+cd ${ALIGNED_DIR}/
+
+## If directory does not exist, create
+mkdir -p ${ALIGNED_DIR}/baseRecalibrate
 
 sampleName=$1
 vcfid=$2
 vcfid="${vcfid%"${vcfid##*[![:space:]]}"}"
```
sof202, line 41: I'm struggling to tell what this is doing. I've taken a row from `sortedBDR/0_metadata/matchedVCFIDs.txt` and tried out this line here using a value from the VCFID column.

However, this doesn't seem to actually do anything to the string. 
To elaborate:
```bash
vcfid=201023670019_R06C02_EX162 # Some VCFID from the file I mentioned
vcfid="${vcfid%"${vcfid##*[![:space:]]}"}"
echo $vcfid # returns 201023670019_R06C02_EX162 (the same value)
```

sof202: I know its not your code, just checking if this matters or not

sequencing/ATACSeq/subScripts/compareBamWithGenotypes.sh
```diff
@@ -43,47 +58,58 @@ then
...
+    rm ${ALIGNED_DIR}/baseRecalibrate/${sampleName}_dedup_rglabelled.bam*
+	  rm ${ALIGNED_DIR}/baseRecalibrate/${sampleName}_recal_data.table
 
+    mkdir -p ${ALIGNED_DIR}/genotypeConcordance
+    
+    ${VERIFYBAMID} --vcf ${SNP} --bam ${ALIGNED_DIR}/baseRecalibrate/${sampleName}_baserecal.bam --out ${ALIGNED_DIR}/genotypeConcordance/${sampleName} --ignoreRG --smID ${vcfid} --self
 else
 
     echo "No genotype data available"
+    echo $sampleName, >> $META_DIR/noVCFfound.txt
```
sof202, line 100: Why is there a comma at the end of each of these lines in the `noVCFfound.txt` file? I guess this file is never being read anywhere (not yet at least), so it doesn't really matter. Just curious

sequencing/ATACSeq/Rscripts/collateSexChecks.r
```diff
@@ -1,34 +1,65 @@
...
+source(configFile)
 
 library(ggplot2)
 library(ggpubr)
 library(tidyverse)
-
-
+library(conflicted)
+conflict_prefer("filter", "dplyr")
+conflict_prefer("lag", "dplyr")
```
sof202, line 29: We could probably avoid this extra dependency in `conflicted` if namespaces are used.

So you could do:
```R
dplyr::filter(...)
dplyr::lag(...)
```

sequencing/ATACSeq/jobSubmission/1_batchRunPreAnalysis.sh
```diff
@@ -45,11 +45,15 @@
 echo Job started on:
 date -u
 
+if [[ $1 == '' ]] || [[ ! -d $1 ]]
```
sof202, line 48: The first check here is redundant I think. If $1 is not supplied then [[ ! -d $1 ]] will return false (as '' is not a directory).

sequencing/ATACSeq/Rscripts/collateSexChecks.r
```diff
@@ -1,34 +1,65 @@
...
   spread(V4, V17)
 
 yPeaksWide = yPeaks %>% 
-	select(V4, V10, V11) %>% 
-  spread(V4, V11)
-  
-xPeaksWide$YTot<-rowSums(yPeaksWide[,-1])
-xPeaksWide$sex<-sampleSheet$gender[match(gsub(".chrX.tn5.tagAlign.gz", "", xPeaksWide$V18), sampleSheet$sampleID)]
+	select(V4, V10, V11) %>%
+ spread(V4, V10)
```
sof202: I realise that there is a lot of columns in xPeaks and yPeaks and this will be a pain, but would it be possible to give names to the columns that are being manipulated here? 

You could probably get away with just giving names to the columns that are being manipulated (`V4`,`V10`,`V11`,*etc*). This should make the code more readable in the future.

sequencing/ATACSeq/Rscripts/collateSexChecks.r
sof202: Also, I don't even see you using `filter` or `lag`, so why use `conflicted` at all?

sequencing/ATACSeq/Rscripts/collateSexChecks.r
```diff
@@ -58,18 +89,19 @@ a2 <- ggplot(xPeaksWide, aes(x=XIST, y=YTot, color=sex)) +
...
 dev.off()
 		  
 ## predict sex
-
+#For Shifted Tag Align mode
 sexPredict<-data.frame("sampleID" = gsub(".chrX.tn5.tagAlign.gz", "", xPeaksWide$V18),"LabelledSex" = xPeaksWide$sex, "FIRREratio" = ratioF > thresF,"XISTratio" = ratioX > thresX)
+
 ## TRUE means male
 sexPredict[sexPredict == TRUE]<-"M"
 sexPredict[sexPredict == FALSE]<-"F"
```
sof202: Just a quick check, should this be:
```R
sexPredict[sexPredict[["LabelledSex]] == TRUE]<-"M"
sexPredict[sexPredict[["LabelledSex]] == FALSE]<-"F"
```

sof202: This only matters if there are values in `sexPredict` outside of the LabelledSex column that are `1`, `TRUE` or `FALSE`. Not sure if this is the case or not.

marinafloresp:
@sof202 Thank you for your comments, I will go over then and make appropriate changes. Some scripts have older parts that might be outdated or non-efficient.

sequencing/ATACSeq/Rscripts/collateSexChecks.r
marinafloresp: I am guessing it is just to avoid the conflict message to pop up

sof202: Nah, if anything it makes more logs than normal. Here's the output:

### test.R
```R
library(dplyr)
library(conflicted)
conflict_prefer("filter", "dplyr")
conflict_prefer("lag", "dplyr")
```

### output
```bash
$ Rscript test.R

  Attaching package: ‘dplyr’
  
  The following objects are masked from ‘package:stats’:
  
      filter, lag
  
  The following objects are masked from ‘package:base’:
  
      intersect, setdiff, setequal, union
  
  [conflicted] Will prefer dplyr::filter over any other package.
  [conflicted] Will prefer dplyr::lag over any other package.
```

marinafloresp: I had several issues with this script, so was probably being used before and I got rid of those functions

sequencing/ATACSeq/Rscripts/collateSexChecks.r
sof202: Ahh, so `FIRREratio` and `XISTratio` are supposed to be columns containing `M` and `F`?

sequencing/ATACSeq/jobSubmission/6_batchRunGenotypeCheck.sh
marinafloresp: This step of the script will only make sense to run if you have already run the COMPARE step, as it is where the results are output. The SWITCH step checks those samples that performed poorly in the COMPARE step and aims to find better genotype matches. That is why it is assumed that the genotypeConcordance directory exists already.

sequencing/ATACSeq/Rscripts/collateSexChecks.r
marinafloresp: Yes, as in  `sexPredict<-data.frame("sampleID" = xPeaksWide$sampleID,"LabelledSex" = xPeaksWide$sex, "FIRREratio" = ratioF > thresF,"XISTratio" = ratioX > thresX)`  FIRREratio and XISTratio will be either TRUE or FALSE

sequencing/ATACSeq/jobSubmission/6_batchRunGenotypeCheck.sh
marinafloresp: I guess it will make more sense to add a header row to samples.txt so that the first sample will be index 1 and not 0. My programming side assumes that the first index is 0 XD

sequencing/ATACSeq/jobSubmission/6_batchRunGenotypeCheck.sh
sof202: Is this in the README?

sequencing/ATACSeq/jobSubmission/6_batchRunGenotypeCheck.sh
sof202: You can do that, sure. But then `${SLURM_ARRAY_TASK_ID}` will need to be `2` to select the first data row in `samples.txt` and `matchedVCFIDs.txt` (as `head -1 $file | tail -1` gets you the first row which will now be the header).

sequencing/ATACSeq/Rscripts/collateSexChecks.r
```diff
@@ -24,39 +24,34 @@ source(configFile)
...
 xPeaks<-read.table(xPeakFile, stringsAsFactors = FALSE)
 yPeaks<-read.table(yPeakFile, stringsAsFactors = FALSE,fill = TRUE)
 
-xPeaks$V18<-basename(xPeaks$V18)
-yPeaks$V11<-basename(yPeaks$V11)
+colnames(xPeaks)[c(4,17,18)] <- c("sex-gene","counts","sampleID")
+colnames(yPeaks)[c(4,10,11)] <- c("peak-name","counts","sampleID")
+
+xPeaks$sampleID<-basename(xPeaks$"sampleID")
+yPeaks$sampleID<-basename(yPeaks$"sampleID")
```
sof202, line 40: This is much clearer now, thank you very much.

sequencing/ATACSeq/Rscripts/collateDataQualityStats.Rmd
```diff
@@ -727,10 +722,11 @@ corMergeStats<-cbind(mergeStats$Total.Sequences,mergeStats$total_deduplicated_pe
                      mergeStats$PBC1,
                      propNucleosomesAll[,c(1,2)],
                      diptestStatsAll[,2],
+                     decreasingProps[,1],
                      fripStats$BAMPeaks,
                      fripStats$FripMACS3PE)
 colnames(corMergeStats)<- c('total\nreads', 'dedup', 'poor\nqual', 'mt\nreads', 'alignt\nrate', 'distinct\nreads', 
-                            'NFR','PBC1',  "PropNFR", "PropMono", "DipP","NPeaks_PE", "FRIP_PE")
+                            'NFR','PBC1',  "PropNFR", "PropMono", "DipP","Periodicity","NPeaks_PE", "FRIP_PE")
```
sof202, line 729: I didn't know you could put escape sequences in `colnames`, thanks Marina

sequencing/ATACSeq/subScripts/alignment.sh
```diff
@@ -73,7 +73,8 @@ samtools stats ${ALIGNED_DIR}/${sampleName}_noMT.bam	> ${ALIGNED_DIR}/QCOutput/$
 # only keep properly paired reads
 echo "filtering aligned reads"
 samtools view -F 524 -f 2  -q 30 -u ${ALIGNED_DIR}/${sampleName}_noMT.bam | samtools sort -n /dev/stdin -o ${ALIGNED_DIR}/${sampleName}_q30.tmp.nmsrt.bam
-samtools view -h ${ALIGNED_DIR}/${sampleName}_q30.tmp.nmsrt.bam | $(which assign_multimappers.py) -k $multimap --paired-end | samtools fixmate -r /dev/stdin ${ALIGNED_DIR}/${sampleName}_q30.tmp.nmsrt.fixmate.bam
+echo $MULTIMAP/assign_multimappers.py
+samtools view -h ${ALIGNED_DIR}/${sampleName}_q30.tmp.nmsrt.bam | $MULTIMAP/assign_multimappers.py -k $multimap --paired-end | samtools fixmate -r /dev/stdin ${ALIGNED_DIR}/${sampleName}_q30.tmp.nmsrt.fixmate.bam
```
sof202, line 77: I forgot that this line requires `assign_multimappers.py` to be on `$PATH`, good find.

ejh243 (approved):
Approving this to merge with main. Any bugs can be fixed as issues if needed.

Issue #166, opened 2024-08-02
[Enhancement suggestion]: Use of renv to manage packages


ejh243:
### Contact Details

e.j.hannon@exeter.ac.uk

### Enhancement suggestion

As discussed in the analysis meeting we will trial using renv to install and manage packages.

This can be segmented into different folders for different parts of the pipeline. I will test initially on DNAm analysis and pre-processing folders.

PR #175, opened 2024-08-23
166 enhancement suggestion use of renv to manage packages

ejh243:
# Description

This pull request will provide files for renv to be used to create reproducible environments for R scripts. There are separate environments for each stage of the pipeline organised by folders. Currently in place for 
* array/DNAm/preprocessing
* array/DNAm/analysis/EWAS
*  array/DNAm/analysis/methodsDevelopment

## Issue ticket number 

This pull request is to address issue: #166 .

## Type of pull request

- [ ] Bug fix
- [X] New feature/enhancement
- [ ] Code refactor
- [X] Documentation update

## Checklist


sof202:
Steps: 
```bash
[sof202@login01 BrainFANS] (renv ✓)$ module load R/4.2.1-foss-2022a                                                                                                                      [sof202@login01 BrainFANS] (renv ✓)$ cd array/DNAm/analysis/methodsDevelopment/
[sof202@login01 methodsDevelopment] (renv ✓)$ R
```

```R
...
# Bootstrapping renv 1.0.7 ---------------------------------------------------
- Downloading renv ... OK
- Installing renv  ... OK

- Project '~/Repositories/BrainFANS/array/DNAm/analysis/methodsDevelopment' loaded. [renv 1.0.7]
- One or more packages recorded in the lockfile are not installed.
- Use `renv::status()` for more details.
Warning message:
renv took longer than expected (20 seconds) to activate the sandbox.

The sandbox can be disabled by setting:

    RENV_CONFIG_SANDBOX_ENABLED = FALSE

within an appropriate start-up .Renviron file.

See `?renv::config` for more details.
```

Sandbox documentation: https://rstudio.github.io/renv/reference/sandbox.html

I then disabled the sandbox mode and the packages download as expected (installation to come).

sof202:
Test list:
[ ] /preprocessing/
[ ] /analysis/EWAS/
[ ] /analysis/methodsDevelopment/

sof202:
> Steps:
> 
> ```shell
> [sof202@login01 BrainFANS] (renv ✓)$ module load R/4.2.1-foss-2022a                                                                                                                      [sof202@login01 BrainFANS] (renv ✓)$ cd array/DNAm/analysis/methodsDevelopment/
> [sof202@login01 methodsDevelopment] (renv ✓)$ R
> ```
> 
> ```r
> ...
> # Bootstrapping renv 1.0.7 ---------------------------------------------------
> - Downloading renv ... OK
> - Installing renv  ... OK
> 
> - Project '~/Repositories/BrainFANS/array/DNAm/analysis/methodsDevelopment' loaded. [renv 1.0.7]
> - One or more packages recorded in the lockfile are not installed.
> - Use `renv::status()` for more details.
> Warning message:
> renv took longer than expected (20 seconds) to activate the sandbox.
> 
> The sandbox can be disabled by setting:
> 
>     RENV_CONFIG_SANDBOX_ENABLED = FALSE
> 
> within an appropriate start-up .Renviron file.
> 
> See `?renv::config` for more details.
> ```
> 
> Sandbox documentation: https://rstudio.github.io/renv/reference/sandbox.html
> 
> I then disabled the sandbox mode and the packages download as expected (installation to come). I am not sure if this is a requirement among users, more people will need to test I guess.

@SiyiSEA got this upon doing the same as me:

# Bootstrapping renv 1.0.7 ---------------------------------------------------
- Downloading renv ... OK
- Installing renv  ... OK

- Project '~/QC/BrainFANS/array/DNAm/analysis/methodsDevelopment' loaded. [renv 1.0.7]
- One or more packages recorded in the lockfile are not installed.
- Use `renv::status()` for more details.

sof202:
As such, it might be a better idea to use `renv::load` in the scripts after all. The behaviour of loading up an R session in the directory where the renv.lock file doesn't seem to be consistent between users.

sof202:
With the Github repositories, it might be a good idea to specify tags or commit hashes instead of HEAD. The pipeline will be a bit more stable that way (we can update these libraries when we need to instead of whenever a commit is made to them)

sof202:
Also, the `.gitignore` files renv creates have not been added to the pull request (these will ensure directories like renv/library/ are not committed to the repository)

sof202:
I've hit a brick wall with Matrix. I expected this to happen but I really can't get around it.

Running 
```R
renv::install("Matrix")
```

gives:
```
Error: package 'Matrix' is not available
Traceback (most recent calls last):
12: renv::install("Matrix")                                                                                                                                                         11: retrieve(packages)                                                                                                                                                              10: handler(package, renv_retrieve_impl(package))                                                                                                                                    9: renv_retrieve_impl(package)                                                                                                                                                      8: withCallingHandlers(renv_available_packages_latest(package),                                                                                                                            error = function(err) stopf("package '%s' is not available",                                                                                                                            package))                                                                                                                                                                7: renv_available_packages_latest(package)                                                                                                                                          6: stopf("package '%s' is not available", package)                                                                                                                                  5: stop(sprintf(fmt, ...), call. = call.)                                                                                                                                           4: .handleSimpleError(function (err)                                                                                                                                                   {                                                                                                                                                                                       ...                                                                                                                                                                             }, "package 'Matrix' is not available", base::quote(NULL))                                                                                                                       3: h(simpleError(msg, call))                                                                                                                                                        2: stopf("package '%s' is not available", package)                                                                                                                                  1: stop(sprintf(fmt, ...), call. = call.)
```

I presume this is because Matrix already exists in the default library (not site packages) and so renv tries to create a symlink between the default cache and the renv/library/ directory. However, this is having a problem for some reason. 

Ideally I would just remove Matrix from the default library as this is in my home directory under sandbox (`$HOME/.cache/R/renv/sandbox/...). If I could just do this then renv wouldn't try to source matrix from my cache and instead install and build it itself. This unfortunately is not possible. The symlink is pointing to a directory I do not have permissions for (the lib64 directory of the R installation on ISCA). Because I don't have permissions for the location of the symlink, I can't use `unlink` or `rm` on the symlink (which seems really silly to me, but I'm sure there is a good reason linux does this).

I also can't turn off SANDBOX in my `.Renviron` file as this results in site packages being used to obtain cached packages. Matrix is also included in site packages and so it ends up being the exact same problem (except now it is the actual installation of Matrix, not a symlink to it).

I have no idea how to get around this. I even tried doing renv's job for it by creating the necessary files, but I guess I don't know everything it requires and so it still complains (also, if I have to go this far to download one package, it completely defeats the purpose of using renv in the first place).

These problems could be avoided with a conda environment.

sof202:
Ok, so with conda environments, I have found that simply using 
```R
renv::install("Matrix")
```

will fail when using a version of <4.4.

If I start using >= 4.4, I no longer have these issues (and I can install the remaining packages).

sof202:
I've managed to get this to work when I use conda to get the latest version of conda. Perhaps `matrix` just doesn't like earlier versions of R.

sof202:
invalidated by #248

Issue #177, opened 2024-08-23
[Bug]: Possible bug in QC.rmd script in DNA microarray QC pipeline


dmsoanes:
### Contact Details

d.m.soanes@exeter.ac.uk

### What happened?

Microarray data was not cell sorted
The first part of the QC pipeline seemed to work fine:
1: Column names were correct
2: raw.gds and rawNorm.gds files were produced
3: calcQCMetrics.r worked (files attached) - although there was a warning: Sentrix position model failed, skipping not sure if that is important
4: The markdown script QC.rmd failed:

### How can the bug be reproduced?

Uploaded config, output, log and sample sheet files - other files can be found at /lustre/projects/Research_Project-MRC190311/DNAm/Darren/OHSU_006PLT_JN
[config.r.txt](https://github.com/user-attachments/files/16727132/config.r.txt)
[config.txt](https://github.com/user-attachments/files/16727134/config.txt)
[PassQCStatusAllSamples.csv](https://github.com/user-attachments/files/16727135/PassQCStatusAllSamples.csv)
[QCDNAdata_2558375.err.txt](https://github.com/user-attachments/files/16727136/QCDNAdata_2558375.err.txt)
[QCDNAdata_2558375.log](https://github.com/user-attachments/files/16727137/QCDNAdata_2558375.log)
[QCMetricsPostSampleCheck.csv](https://github.com/user-attachments/files/16727138/QCMetricsPostSampleCheck.csv)
[sampleSheet.csv](https://github.com/user-attachments/files/16727139/sampleSheet.csv)

### Relevant log output

```shell
processing file: QC.rmd
Error in `contrasts<-`:
! contrasts can be applied only to factors with 2 or more levels
Backtrace:
 1. stats::lm(...)
 3. stats::model.matrix.default(mt, mf, contrasts)
 4. stats::`contrasts<-`(`*tmp*`, value = contr.funs[1 + isOF[nn]])
Warning message:
In in_dir(input_dir(), expr) :
  You changed the working directory to /lustre/projects/Research_Project-MRC190311/DNAm/Darren/OHSU_006PLT_JN (probably via setwd()). It will be restored to /lustre/projects/Research_Project-MRC190311/DNAm/Darren/OHSU_006PLT_JN/original_scripts/BrainFANS/array/DNAm/preprocessing. See the Note section in ?knitr::knit

Quitting from lines 402-413 [controlProbeHeatmap] (QC.rmd)
Execution halted
```

[...]

Issue #173, opened 2024-08-21
[Bug]: Possible bug in QC.rmd script in DNAm QC pipeline 


dmsoanes:
### Contact Details

d.m.soanes@exeter.ac.uk

### What happened?

Microarray data was not cell sorted (small test dataset)
The first part of the QC pipeline seemed to work fine:
1: Column names were correct
2: raw.gds and rawNorm.gds files were produced
3: calcQCMetrics.r worked (files attached) - although there was a warning: Sentrix position model failed, skipping not sure if that is important
4: The markdown script QC.rmd failed:

### How can the bug be reproduced?

Uploaded config, output, log and sample sheet files - other files can be found at /lustre/projects/Research_Project-MRC190311/DNAm/Darren/OHSU_006PLT_JN
[config.txt](https://github.com/user-attachments/files/16689449/config.txt)
[PassQCStatusAllSamples.csv](https://github.com/user-attachments/files/16689450/PassQCStatusAllSamples.csv)
[QCDNAdata_2556271.log](https://github.com/user-attachments/files/16689451/QCDNAdata_2556271.log)
[QCMetricsPostSampleCheck.csv](https://github.com/user-attachments/files/16689452/QCMetricsPostSampleCheck.csv)
[sampleSheet.csv](https://github.com/user-attachments/files/16689453/sampleSheet.csv)
[config.r.txt](https://github.com/user-attachments/files/16689491/config.r.txt)
[QCDNAdata_2556271.err.txt](https://github.com/user-attachments/files/16689492/QCDNAdata_2556271.err.txt)

### Relevant log output

```shell
processing file: QC.rmd
Error in `names(x) <- value`:
! 'names' attribute [6] must be the same length as the vector [5]
Backtrace:
 1. base::`colnames<-`(`*tmp*`, value = `<chr>`)
 2. base::`colnames<-`(`*tmp*`, value = `<chr>`)
Warning message:
In in_dir(input_dir(), expr) :
  You changed the working directory to /lustre/projects/Research_Project-MRC190311/DNAm/Darren/OHSU_006PLT_JN (probably via setwd()). It will be restored to /lustre/projects/Research_Project-MRC190311/DNAm/Darren/OHSU_006PLT_JN/original_scripts/BrainFANS/array/DNAm/preprocessing. See the Note section in ?knitr::knit

Quitting from lines 61-141 [definPlotParams] (QC.rmd)
Execution halted
```

sof202:
[...]

Issue #33, opened 2024-02-06
[Enhancement suggestion]: create basic technical QC or 'full' QC options for running DNAm QC pipeline


ew367:
### Enhancement suggestion

Have different options for running a quicker version of the QC or full version of the QC

**quicker technical only version**: intensities/bscon/PCA of control-probe intensities/detection p value/count number of missing values
e.g. for collaborators data we are not involved in analysing/initally checking that the experiment as worked
**outputs**: html QC report, csv file of QC metrics

**Full version:** run above and then additionally run: Horvaths epigenetic age/Cortical Clock Age/check effect of normalisation/predict sex/check duplicate samples using SNPs on array/compare to external snp data
**outputs:** as above, plus: updated gds file containing normalised beta values, QC rdata object containing QCmetrics, snpCor, betas.pca, ctrl.pca, pFOut, geno.mat, betas.rs, rsbetas, samplesPassFail csv file, html QC report

ejh243:
Technical QC is effectively just stage 1 of the QC, i.e. no sexCheck, no genoCheck, no ctCheck. 

Could add variable to config file to specify technical QC only which would then override config specified parameters.

Also need to prevent execution of scripts post calc QC metrics.

Do we need a separate rmarkdown for this?

sof202:
#216?

ew367:
Yes, I think this address this issue

PR #178, opened 2024-08-23
refactor: convert section "definPlotParams" into functions

sof202:
# Description

This pull request:

- Forcibly stops the creation of the QCmetrics report if columns are not identical between the config file and the sampleSheet.csv. The error message is now also updated such that it provides the necessary information to fix the issue in the error logs.
- Correctly identifies columns that should not be brought forward into the later plots 
  - Which include:
    - Columns with all missing values
    - Columns with only a single unique value (`contrasts()` will fail in this case)
    - Columns that contain all unique values (not sure why this is the case, but the old code did this)
- Refactors the definPlotParams section to use functions and less repeated code

## Issue ticket number

This pull request is to address issues: #173 #177 (maybe #33 ?).

## Type of pull request

- [x] Bug fix
- [ ] New feature/enhancement
- [x] Code refactor
- [ ] Documentation update

## Checklist

### DIFF
--- a/array/DNAm/preprocessing/QC.rmd
+++ b/array/DNAm/preprocessing/QC.rmd
@@ -93,15 +93,19 @@ is_wanted_column <- function(column_name, QCmetrics) {
   ))
 }
 
-define_plot_parameters <- function(QCmetrics, techVar, bioVar) {
-  potential_plot_columns <- c(techVar, bioVar)
+remove_unwanted_columns <- function(columns, QCmetrics) {
   columns_to_keep <- vapply(
-    potential_plot_columns,
+    columns,
     is_wanted_column,
     logical(1),
     QCmetrics
   )
-  plot_columns <- potential_plot_columns[columns_to_keep]
+  return(columns[columns_to_keep])
+}
+
+define_plot_parameters <- function(QCmetrics, techVar, bioVar) {
+  potential_plot_columns <- c(techVar, bioVar)
+  plot_columns <- remove_unwanted_columns(potential_plot_columns)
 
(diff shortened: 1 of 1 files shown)


sof202:
Current problems with my code:
If both `techVar` and `bioVar` are undefined, then `plotCols` will just be the `Basename` column of `QCmetrics`. Is this the desired effect? 
(This is what the code did before, just checking if this is actually what we want.)

sof202:
> # Current problems with my code
> ## One
> If both `techVar` and `bioVar` are undefined, then `plotCols` will just be the `Basename` column of `QCmetrics`. Is this the desired effect? (This is what the code did before, just checking if this is actually what we want.)
> 
> ## Two
> Not sure if I like the if statement hell:
> 
> ```r
> if (all(is.na(column))) return(FALSE)
> if (length(unique(column)) == 1L) return(FALSE)
> if (anyDuplicated(column) == 0L) return(FALSE)
> return(TRUE)
> ```
> 
> I could change this to:
> 
> ```r
> if (
> all(is.na(column)) || 
> length(unique(column)) == 1L || 
> anyDuplicated(column) == 0L) {
>   return(FALSE)
> } 
> return(TRUE)
> ```
> 
> Or:
> 
> ```r
> return(!(all(is.na(column)) || 
>           length(unique(column)) == 1L || 
>           anyDuplicated(column) == 0L))
> ```
> 
> Though I'm not convinced these are much cleaner. I'm open to suggestions

I've chosen the 3rd option now

SiyiSEA (approved):

Issue #184, opened 2024-08-30
[Enhancement suggestion]: Check variables exist before execution


ejh243:
### Contact Details

e.j.hannon@exeter.ac.uk

### Enhancement suggestion

At the top of scripts that depend on variables defined in the config files, to check they exist and print helpful error messages if they do not.

ew367:
Resolvedby adding `Rscript checkRconfigFile.r ${DATADIR}` to the pipeline

PR #208, opened 2024-09-17
184 check r config

ew367:
# Description

This PR will...

- update config.example.file

- add script to check necessary config file params are present and in correct format and add the script call to job submission script

- update cetygo and clusterCellTypes script to use ctCheck param instead of cellSorted

FYI

changes to config.example.r

- removes redundant variables from config file

- repo search for arrayVersion, otherCT, pnthres and perc returns no results other than config file itself

- cellSorted is redundant as is essentially doing same as ctCheck = T/F

- variables now grouped according to where/if they are used

## Issue ticket number #184 

This pull request is to address issue: #184 

## Type of pull request

- [ ] Bug fix
- [x] New feature/enhancement
- [x] Code refactor
- [ ] Documentation update

## Checklist

### DIFF
--- a/array/DNAm/config/config.example.r
+++ b/array/DNAm/config/config.example.r
@@ -1,33 +1,32 @@
 ## Study parameters
 projectTitle<-"MRC Schizophrenia FANS samples"
 processedBy<-"Complex Disease Epigenetic Group, University of Exeter Medical School"
-arrayVersion<-"Illumina EPIC microarray"
 tissueType<-"brain" # needs to be brain or blood to generate relevant cell composition estimates
-cellSorted<-FALSE # turns off cell type clustering check
+
 
 ## technical variables
-
 techVar<-c("Sentrix_ID","Sentrix_Position")
 
+
 ## biological variables
-
 bioVar<-c("Individual_ID","Cell_Type","Sex")
-
-predDistinctCT<-c("NeuN+", "Sox10+", "IRF8+", "Triple-", "Total")
-neunCT<-c("NeuN+", "SATB2+")
-otherCT<-c("Double-", "IRF8+", "SATB2", "Sox10+", "Triple-")
-
 
 
 ## QC thresholds
 thresBS<-80
 intenThres<-500
 nvThres<-0.1
+perMiss<-2
+
+
+## Optional QC steps
 sexCheck<-TRUE
 snpCheck<-TRUE
 ctCheck<-TRUE
-perMiss<-2
+
+
+## ctCheck variables
+predDistinctCT<-c("NeuN+", "Sox10+", "IRF8+", "Triple-", "Total")
+neunCT<-c("NeuN+", "SATB2+")
 studentThres<-1.5
 nSDThres<-3
-pnthres<-0.05
-perc<-1
--- a/array/DNAm/config/config.example.r
+++ b/array/DNAm/config/config.example.r
@@ -2,6 +2,7 @@
 projectTitle<-"MRC Schizophrenia FANS samples"
 processedBy<-"Complex Disease Epigenetic Group, University of Exeter Medical School"
 tissueType<-"brain" # needs to be brain or blood to generate relevant cell composition estimates
+arrayType<-"V2" # must be one of "450K" / "V1" / "V2"
 
 
 ## technical variables
--- a/array/DNAm/preprocessing/checkRconfigFile.r
+++ b/array/DNAm/preprocessing/checkRconfigFile.r
@@ -1,0 +1,161 @@
+##---------------------------------------------------------------------#
+##
+## Title: Check config.r 
+##
+## Purpose of script: Check config file parameters are present and in correct format
+##                    and if not exit the script with error message
+##
+##---------------------------------------------------------------------#
+
+
+#----------------------------------------------------------------------#
+# NOTES
+#----------------------------------------------------------------------#
+
+# Project Folder is provided on the command line
+
+
+#----------------------------------------------------------------------#
+# DEFINE PARAMETERS
+#----------------------------------------------------------------------#
+
+print("checking config.r file parameters are present and correctly formatted...")
+
+'%ni%' <- Negate('%in%') # define '%ni%' (not in)
+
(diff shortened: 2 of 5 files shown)


array/DNAm/preprocessing/jobSubmission/1_runDNAmQC.sh
```diff
@@ -34,6 +34,8 @@ module load $RVERS   # load specified R version
 
 cd ${SCRIPTSDIR}/array/DNAm/preprocessing/
 
+Rscript checkRconfigFile.r ${DATADIR}
```
sof202, line 37: I haven't read through this file properly yet, but it looks as though if the config file is malformed, the following commands in the jobsubmission shell script will run anyways.

Although the addition of the config checker is a great step, it would be great if execution of the entire shell script stops in such cases (as otherwise, the user might not know about these errors if they don't check the logs)

sof202: I have an idea of how to do this, which I will try to implement tomorrow.

array/DNAm/config/config.example.r
```diff
@@ -1,34 +1,33 @@
...
 snpCheck<-TRUE
 ctCheck<-TRUE
-perMiss<-2
+
+
+## ctCheck variables
+predDistinctCT<-c("NeuN+", "Sox10+", "IRF8+", "Triple-", "Total")
+neunCT<-c("NeuN+", "SATB2+")
 studentThres<-1.5
 nSDThres<-3
```
sof202, line 33: Happy that this file is nicely formatted now. Thank you

array/DNAm/preprocessing/checkRconfigFile.r
```diff
@@ -0,0 +1,164 @@
...
+} 
+
+                    
+
+# if ctCheck is TRUE, check extra params and thresholds needed to run clusterCellTypes.r script
+
+
+if(ctCheck == TRUE){
+  for(i in ctCellParams){
+    if(exists(i)){
```
sof202, line 116: We've already checked if these exist, so this isn't required.

array/DNAm/preprocessing/checkRconfigFile.r
```diff
@@ -0,0 +1,164 @@
...
+# Project Folder is provided on the command line
+
+
+#----------------------------------------------------------------------#
+# DEFINE PARAMETERS
+#----------------------------------------------------------------------#
+
+print("checking config.r file parameters are present and correctly formatted...")
+
+'%ni%' <- Negate('%in%') # define '%ni%' (not in)
```
sof202, line 24: This is never used in the file?

array/DNAm/preprocessing/checkRconfigFile.r
```diff
@@ -0,0 +1,164 @@
...
+    stop(paste0("'", i, "' must be included in techVar"))
+  }
+}
+
+
+for(i in c("Sentrix_ID", "Sentrix_Position")){
+  if(!i %in% techVar){
+    stop(paste0("'", i, "' must be included in techVar"))
+  }
+}
```
sof202, line 159: Repeated code. I presume one of these is supposed to be for bioVar.

array/DNAm/preprocessing/checkRconfigFile.r
```diff
@@ -0,0 +1,164 @@
...
+#----------------------------------------------------------------------#
+# Check config file parameters
+#----------------------------------------------------------------------#
+
+
+# check necessary parameters exist
+
+for(i in c(qcRmdParams, qcthres, logicalParams, ungrouped)){
+  if(!exists(i)){
+    stop(paste0("'", i, "' must be defined in the config file" ))
```
sof202, line 75: There is a small problem with using `stop` here. The user will have to run this script checker way more times than necessary as the script stops checking once it finds a single error. If there are 5 errors, then the user will have to run this script 6 times minimum.

array/DNAm/preprocessing/jobSubmission/1_runDNAmQC.sh
sof202: done.

array/DNAm/preprocessing/checkRconfigFile.r
ew367: good point - leftover code and corrected in #209

array/DNAm/preprocessing/checkRconfigFile.r
ew367: good point - corrected in #209

array/DNAm/preprocessing/jobSubmission/1_runDNAmQC.sh
ew367: It would be nice to add this after the checkColnamesSampleSheet.r script too

array/DNAm/preprocessing/checkRconfigFile.r
ew367: good point - corrected in #209

ew367:
@sof202 is there anything outstanding on this before it can be merged?

array/DNAm/preprocessing/jobSubmission/1_runDNAmQC.sh
sof202: I've created a issue for this

sof202:
I guess not. Obviously we will have to revisit this script in the future. But for now it is fine. We said how we want to hard code certain values into `techVar`/`bioVar`, but those haven't been implemented yet (so there is no point in accounting for them in this PR)

sof202 (approved):

ew367:
> I guess not. Obviously we will have to revisit this script in the future. But for now it is fine. We said how we want to hard code certain values into `techVar`/`bioVar`, but those haven't been implemented yet (so there is no point in accounting for them in this PR)

I'll open a seperate issue for this

sof202:
Ok, now I'm happy.

ew367 (commented):
Great

PR #209, opened 2024-09-18
refactor: checkRconfig

sof202:
# Description

This pull request will refactor the code in #208. Obviously not everything needs to be carried across. #208 works in its own right, there are just a few quality of life additions I thought to add.

The main QOLs are:
- Warning messages used instead of stop messages so all bad parameters can be identified in one run
- Use of a function to check parameters existence and type/class (for future use)
- Error messages (relating to bad parameters) are separated with newlines for readability in log files.

## Type of pull request

- [ ] Bug fix
- [ ] New feature/enhancement
- [x] Code refactor
- [ ] Documentation update

## Checklist

### DIFF
--- a/array/DNAm/preprocessing/checkRconfigFile.r
+++ b/array/DNAm/preprocessing/checkRconfigFile.r
@@ -41,14 +41,17 @@ ctCellParams <- c("predDistinctCT", "neunCT")
 #----------------------------------------------------------------------#
 # Check config file parameters
 #----------------------------------------------------------------------#
+bad_parameter <- FALSE
 
 check_parameters <- function(parameters, type_check, stop_message) {
   for (parameter in parameters) {
     if (!exists(parameter)) {
-      stop("'", parameter, "' must be defined in the config file")
+      bad_parameter <- TRUE
+      warning("'", parameter, "' must be defined in the config file")
     }
     if (!type_check(get(parameter))) {
-      stop("'", parameter, "' ", stop_message)
+      bad_parameter <- TRUE
+      warning("'", parameter, "' ", stop_message)
     }
     message("'", parameter, "' is correctly defined.")
   }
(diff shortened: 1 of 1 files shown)


array/DNAm/preprocessing/checkRconfigFile.r
```diff
@@ -21,144 +21,84 @@
...
+if (!toupper(arrayType) %in% c("HM450K", "V1", "V2")) {
+  bad_parameter_exists <- TRUE
+  warning("\nUnrecognised arrayType. Must be 'HM450K', 'V1' or 'V2'\n")
 }
 
 
-for(i in c("Sentrix_ID", "Sentrix_Position")){
-  if(!i %in% techVar){
-    stop(paste0("'", i, "' must be included in techVar"))
+for (i in c("Individual_ID", "Cell_Type", "Sex")) {
```
ew367, line 83: Technically I think the the scripts can handle bioVar and techVar not being defined (line 133 and 134 of QC.rmd: 

`if (!exists("techVar")) techVar <- NULL`
`if (!exists("bioVar")) bioVar <- NULL`

In practice. Sentrix_ID and Sentrix_Position are always going to need to be included in the QC. It would probably be better to hard code them in, but that's beyond the scope of this PR..

array/DNAm/preprocessing/checkRconfigFile.r
```diff
@@ -21,144 +21,84 @@
...
+  if (!i %in% bioVar) {
+    bad_parameter_exists <- TRUE
+    warning("\n'", i, "' must be included in bioVar\n")
   }
 }
 
-
-for(i in c("Individual_ID","Cell_Type","Sex")){
```
ew367, line 155: Cell_Type is not going to exist for bulk datasets

ew367 (approved):

array/DNAm/preprocessing/checkRconfigFile.r
sof202: Ah ok, I was just copying and pasting. I thought cell type would be just NAs, but still needs to be in bioVar. I can fix this.

Issue #213, opened 2024-09-20
[Enhancement suggestion]: Add stops on unsuccessful scripts


sof202:
### Enhancement suggestion

For unsuccessful scripts in `1_runDNAmQC`, a helpful log message should be displayed before exiting the script entirely. This is to stop the shell script from running when we know for a fact it will end up failing.

We could do this with `set -e`, but then log messages couldn't be displayed to help users identify why the script ended.

PR #215, opened 2024-09-20
feat: Add more error logging to DNAm pipeline

sof202:
# Description

This pull request will add more in depth error logging to the main DNAm job submission script. If an R script fails that is needed for subsequent steps (for example if the sample sheet is in the incorrect format), then a helpful error message will be written to the error logs (and the script will exit early with exit code 1).

## Example
The error logs will look like this:

image

## Issue ticket number

This pull request is to address issue: #213.

## Type of pull request

- [ ] Bug fix
- [x] New feature/enhancement
- [ ] Code refactor
- [ ] Documentation update

## Checklist

### DIFF
--- a/array/DNAm/preprocessing/jobSubmission/1_runDNAmQC.sh
+++ b/array/DNAm/preprocessing/jobSubmission/1_runDNAmQC.sh
@@ -38,7 +38,8 @@ date -u
 echo Log files intially stored in: ${SLURM_SUBMIT_DIR}/QCDNAdata_${SLURM_JOB_ID}.log and ${SLURM_SUBMIT_DIR}/QCDNAdata_${SLURM_JOB_ID}.err
 
 config_file=$1
-source "${config_file}" || print_error_message \ "The provided config file was not sourced correctly." \
+source "${config_file}" || print_error_message \
+    "The provided config file was not sourced correctly." \
     "Please check the path you gave exists, exiting..." 
 
 echo "Processing data located in :" ${DATADIR}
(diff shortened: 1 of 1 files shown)


sof202:
There is a potential problem here however. I cannot install the packages required for BrainFANS due to `matrix` problems (*etc.*).

It shouldn't be the case, but could someone check that the error message for installing the R libraries doesn't appear upon successfully installing the R packages? I'm not 100% sure on how R handles error codes when installing packages. There's a chance that R might output an error code of something other than 0 or 1 (which will still flag the exiting of the script) if the packages are already installed.

sof202:
> There is a potential problem here however. I cannot install the packages required for BrainFANS due to `matrix` problems (_etc._).
> 
> It shouldn't be the case, but could someone check that the error message for installing the R libraries doesn't appear upon successfully installing the R packages? I'm not 100% sure on how R handles error codes when installing packages. There's a chance that R might output an error code of something other than 0 or 1 (which will still flag the exiting of the script) if the packages are already installed.

Just checked this, the error code for installing a package that has already been installed is still 0. No problem here.

sof202:
Only problem I encountered is that if a package cannot be found (for example, the version of R being used is too old), R doesn't return an error/warning message so this approach won't work.

```text
> install.packages("this-doesnt-exist")
Installing package into ‘.../R’
(as ‘lib’ is unspecified)
Warning message:
package ‘this-doesnt-exist’ is not available for this version of R

A version of this package for your version of R might be available elsewhere,
see the ideas at
https://cran.r-project.org/doc/manuals/r-patched/R-admin.html#Installing-packages
```

```bash
echo $?
## 0
```

ejh243 (approved):
Nice additions

Issue #211, opened 2024-09-20
[Enhancement suggestion]: QC report add document code version 


ejh243:
### Contact Details

e.j.hannon@exeter.ac.uk

### Enhancement suggestion

In QC report add a statement specifying which version of the repo was used and the last commit.

PR #217, opened 2024-09-20
Add repository version to QC rmarkdown

sof202:
# Description

This pull request will add this line:

image

To the QC.html report under the "Study Information" field.

## Issue ticket number

This pull request is to address issue: #211.

## Type of pull request

- [ ] Bug fix
- [x] New feature/enhancement
- [ ] Code refactor
- [ ] Documentation update

## Checklist

### DIFF
--- a/array/DNAm/preprocessing/QC.rmd
+++ b/array/DNAm/preprocessing/QC.rmd
@@ -27,9 +27,8 @@ library(data.table)
 source(args[4]) ### change the content of this file to run QC with different thresholds
 ### prior to running this Rmarkdown which summarises the QC output, QC metrics must have been generated
 
-
-refDir <- args[3]
 dataDir <- args[2]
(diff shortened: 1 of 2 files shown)


sof202:
## Problem
This does require the user to have Git installed on their system and for git to be on the path. This should be a given in all honesty on the vast majority of systems, but the code isn't 100% portable.

## Solutions
- We could add another module call to the config file
- We could add a 'software requirements' section to the README (this way we can specify all software that we are assuming the user has installed, like bash, SLURM and GNU core utilities)

ejh243 (approved):
Do we also want to log which branch - although I assume we can work out from commit.

sof202:
Commit is more useful I reckon. Branches get deleted and can be renamed at a whim, not particularly useful information unless we had a long standing branch (like `dev` for example).

Issue #214, opened 2024-09-20
[Bug]: QC Report Section 4.2 reconfirming genetic relationships


Molly-Endicott:
### Contact Details

me460@exeter.ac.uk

### What happened?

When generating the QC.html, we think the table in section 4.2 "Reconfirming genetic relationships are correct" is identifying the samples which have been filtered out during QC as samples from the same individual. The genetically identical samples are correctly identified in section 3.1 but an additional column appears postQC in section 4.2 and corresponds to the number of samples that have been removed during QC in my dataset. 
![Screenshot 2024-09-20 at 11 50 20](https://github.com/user-attachments/assets/9757bffa-2769-4707-b5b2-1e5b1be3807d)
![Screenshot 2024-09-20 at 11 50 51](https://github.com/user-attachments/assets/be39674c-4b20-4810-9ac1-e9d552f4901f)

### How can the bug be reproduced?

The data I am using to generate the QC.html is not currently publicly available. However, please just let me know if I can provide any further information.

PR #218, opened 2024-09-20
fix: predictedduplicates containing empty strings

sof202:
# Description

This pull request will remove samples (from QC) would have "" in their spot in predictedduplicates. These are then clustered together in the table call below, resulting in section 4.2 giving a confusing result (saying these samples belong to the same individual despite being removed). This no longer occurs.

## Issue ticket number

This pull request is to address issue: #214.

## Type of pull request

- [x] Bug fix
- [ ] New feature/enhancement
- [ ] Code refactor
- [ ] Documentation update

## Checklist


sof202:
## Potential refactor

I could implement this fix into the function that created `predictedduplicates` instead. I decided to put the fix next to where the error was happening so that the accompanying explanation (comment) has more context.

Happy to move it into the function (`findDuplicateSamples`) if people believe that is a better choice.

ew367 (approved):
This works

Issue #216, opened 2024-09-20
[Enhancement suggestion]: Remove techVar and bioVar variables and replace with hardcoded Sentrix_ID and Sentrix_Position and optional project variables (projVar)


ew367:
### Enhancement suggestion

Remove techVar and bioVar variables and replace with hardcoded Sentrix_ID and Sentrix_Position and optional project variables (projVar)

Sentrix_ID and Sentrix_Position will need to be included as plot variables in the QC report for all projects so should be hardcoded in to the QC.rmd script. There is also no need to distinguish between technical and biological variables so these can be combined into an overall (optional) project variable.

affected scripts:

config.example.r,  checkRconfigFile.r and QC.rmd files

ew367:
This also address #188

sof202:
Also addresses #33 I think

Issue #188, opened 2024-08-30
[Bug]: technical variables summary never shows


sof202:
### What happened?

In the QC.rmd, Section 1.1 will never show summaries of samples categorised by [insert `techVar` item here]

### How can the bug be reproduced?

Run the QC pipeline and look at QC.html. Section 1.1 will always be empty.

sof202:
## Why?

The code associated with `techVar` is different from `bioVar`:

`techVar`:

```R
for(item in techVar){
  tabTemp<-as.data.frame(table(QCmetrics[,item]))
  if(nrow(tabTemp) < 2){
    pander(
      t(tabTemp),
      caption = paste("Summary of samples categorised by", item),
      row.names = c("Variable", "Number of Samples")
    )
  }
}
```

`bioVar`:

```R
for(item in bioVar){
  pander(
    as.data.frame(table(QCmetrics[, item])),
    caption = paste("Summary of samples categorised by", item),
    col.names = c("Variable", "Number of Samples")
  )
}
```

-----------------

## Problem

Notice how `techVar`'s section of code is almost identical, but it requires `tabTemp` to have less than 2 rows. The only way for this code to render anything on QC.html therefore is for `tabTemp` to have 1 row. However, the only way for `tabTemp` to have a single row is if every value in the column extracted from `QCmetrics` is the same. This is where the problem lies, variables are removed from analysis if all values are the same now thanks to PR #178. As such, `techVar` does not include any columns that can spit out a `tabTemp` that satisfies this `if` statement (and even if it did, such a table would be rather useless).

I presume that this check was implemented to **not** show variables in `techVar` that had all values the same originally (though this would require the `<` to be converted into `>`). But this still doesn't explain why `bioVar` is not handled in this way.

------------------

## Solutions

We can either:

- Remove the conditional check in the `techVar` block
- Convert the `<` to a `>` in the conditional check for the `techVar` block
- Convert the `<` to a `>` in the conditional check for the `techVar` block and add it to the `bioVar` block

sof202:
This will be fixed by the PR addressing #216. Closing the issue.

PR #220, opened 2024-09-23
Remove tech var and bio var variables

ew367:
# Description

This pull request will remove the techVar and bioVar variables from the config.example.r,  checkRconfigFile.r and QC.rmd files and replace them with an overal projVar (project variable).

It will also hard code Sentrix_ID and Sentrix_Position into the QC.rmd doc plots.

## Issue ticket number

This pull request is to address issues: #33, #188 and #216

## Type of pull request

- [x] Bug fix
- [x] New feature/enhancement
- [ ] Code refactor
- [ ] Documentation update

## Checklist

### DIFF
--- a/array/DNAm/preprocessing/QC.rmd
+++ b/array/DNAm/preprocessing/QC.rmd
@@ -40,7 +40,6 @@ if(!"Cell_Type" %in% colnames(QCmetrics)){
 }
 
 
-nSamAll<-nrow(QCmetrics)
 nSamIP<-sum(QCmetrics$intensPASS)
 
 par(mar = c(4,4,1,1))
--- a/array/DNAm/preprocessing/QC.rmd
+++ b/array/DNAm/preprocessing/QC.rmd
@@ -110,7 +110,9 @@ remove_unwanted_columns <- function(columns, QCmetrics) {
 define_plot_parameters <- function(QCmetrics, projVar) {
   plot_columns <- remove_unwanted_columns(projVar, QCmetrics)
 
-  plotCols <- data.frame("Basename" = QCmetrics[["Basename"]])
+  plotCols <- data.frame("Basename" = QCmetrics[["Basename"]],
+                        "Sentrix_ID" = substr(QCmetrics[["Basename"]], 1, 12),
+                        "Sentrix_Position" = substr(QCmetrics[["Basename"]], 14, 19))
   legendParams <- list()
 
   for (column_name in plot_columns) {
--- a/array/DNAm/preprocessing/QC.rmd
+++ b/array/DNAm/preprocessing/QC.rmd
@@ -27,6 +27,7 @@ library(data.table)
 source(args[4]) ### change the content of this file to run QC with different thresholds
 ### prior to running this Rmarkdown which summarises the QC output, QC metrics must have been generated
 
+
 refDir <- args[3]
 dataDir <- args[2]
 setwd(dataDir) 
@@ -38,6 +39,10 @@ load(qcData)
 if(!"Cell_Type" %in% colnames(QCmetrics)){
   QCmetrics$Cell_Type <- NA
 }
+
+# extract Sentrix_ID and _Position from basename as excel can sometimes cause them to be in sci not
+QCmetrics[["Sentrix_ID"]] <- substr(QCmetrics[["Basename"]], 1, 12)
+QCmetrics[["Sentrix_Position"]] = substr(QCmetrics[["Basename"]], 14, 19)
 
 
 nSamIP<-sum(QCmetrics$intensPASS)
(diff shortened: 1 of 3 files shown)


array/DNAm/preprocessing/checkRconfigFile.r
```diff
@@ -80,24 +80,9 @@ if (!toupper(arrayType) %in% c("HM450K", "V1", "V2")) {
...
-
-for (i in c("Individual_ID", "Sex")) {
-  if (!i %in% bioVar) {
-    bad_parameter_exists <- TRUE
-    warning("\n'", i, "' must be included in bioVar\n")
-  }
-}
-
-if (!"Cell_Type" %in% bioVar && ctCheck) {
+if (!"Cell_Type" %in% projVar && ctCheck) {
```
sof202, line 83: A possible problem here is that someone's data might have cell type data, but it is in a column called "cell_type" or "cellType" *etc.*

Not related to this PR though. Just wanted to mention it

array/DNAm/preprocessing/QC.rmd
```diff
@@ -108,11 +112,11 @@ remove_unwanted_columns <- function(columns, QCmetrics) {
   return(columns[columns_to_keep])
 }
 
-define_plot_parameters <- function(QCmetrics, techVar, bioVar) {
-  potential_plot_columns <- c(techVar, bioVar)
-  plot_columns <- remove_unwanted_columns(potential_plot_columns, QCmetrics)
+define_plot_parameters <- function(QCmetrics, projVar) {
+  plot_columns <- c(remove_unwanted_columns(projVar, QCmetrics), "Sentrix_ID", "Sentrix_Position")
```
sof202, line 116: Is there any chance that `Sentrix_ID` or `Sentrix_Position` could be columns that would be removed by `remove_unwanted_columns`? I'm pretty sure the only problem could come from having the same value in every row for these columns (which I doubt will ever happen). 
Should be fine.

sof202:
As long as the stuff I've pointed out aren't problems. This is functional and is an improvement over using `bioVar` and `techVar`.

array/DNAm/preprocessing/QC.rmd
sof202: They could probably be all different values (all entries differ) in some datasets (which would mean they would have previously been removed). This check in `remove_unwanted_columns` was put in place to avoid having variables plotted that don't give the user any extra information.

But I think you said you always want these two on the plots anyways, so this shouldn't matter.

sof202:
This also fixes a problem I had with the `define_plot_parameters` function where it would output blank vectors if `bioVar` and `techVar` were empty. Thank you

array/DNAm/preprocessing/checkRconfigFile.r
ew367: This looks to already be covered by the checkColnamesSampleSheet.r script

```
cond_cols <- c('Sex','Genotype_IID','Cell_Type')

#3. Check conditional column names ---------------------------------------------------------------------------
cat("Sourcing conditional variables from config.r",'\n')
source(configFile)
cond_status <- c(sexCheck,snpCheck,ctCheck) # T/Fs from config file
cat(paste0(c("sexCheck","snpCheck","ctCheck"),"=", cond_status),'\n')

# subset conditional colnames to those TRUE in config
cond_cols.filtered <- cond_cols[cond_status]

if(all(cond_status==F)){
	cat("No conditional variables to check")
}else{
	cat("3. Checking conditional column names: ",'\n')
	cat(cond_cols.filtered,'\n')
	checkColnames(sampleSheet, cond_cols.filtered, type='Conditional')
}
```

array/DNAm/preprocessing/QC.rmd
ew367: Agreed

array/DNAm/preprocessing/QC.rmd
```diff
@@ -39,8 +40,11 @@ if(!"Cell_Type" %in% colnames(QCmetrics)){
   QCmetrics$Cell_Type <- NA
 }
 
+# extract Sentrix_ID and _Position from basename as excel can sometimes cause them to be in sci not
+QCmetrics[["Sentrix_ID"]] <- substr(QCmetrics[["Basename"]], 1, 12)
+QCmetrics[["Sentrix_Position"]] = substr(QCmetrics[["Basename"]], 14, 19)
```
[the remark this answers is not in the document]
ew367: Yes

array/DNAm/preprocessing/checkRconfigFile.r
sof202: Fantastic

array/DNAm/preprocessing/QC.rmd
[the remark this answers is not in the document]
sof202: Cool coolcoolcool

sof202 (approved):

PR #234, opened 2024-10-25
add script to filter out sex specific probes

ew367:
# Description

This pull request will add a standalone script to filter to autosomal probes.

It adds a extra layer to the gds file with a betas matrix for just the autosomal probes, and also saves the matrix and QCmetrics file as a standalone object in the the 3_normalised directory.

## Issue ticket number : none

This pull request is to address issue: #number. - none

## Type of pull request

- [ ] Bug fix
- [x] New feature/enhancement
- [ ] Code refactor
- [ ] Documentation update

## Checklist

### DIFF
--- a/array/DNAm/preprocessing/filterToAutosomes.r
+++ b/array/DNAm/preprocessing/filterToAutosomes.r
@@ -1,0 +1,110 @@
+##---------------------------------------------------------------------#
+##
+## Title: Remove Sex Specific Probes
+##
+##
+##---------------------------------------------------------------------#
+
+
+#----------------------------------------------------------------------#
+# NOTES
+#----------------------------------------------------------------------#
+# project folder is provided on command line
+# ref directory is provided on command line
+# manifest file for epic data is already available in the gds object
+
+
+#----------------------------------------------------------------------#
+# DEFINE PARAMETERS
+#----------------------------------------------------------------------#
+
+args<-commandArgs(trailingOnly = TRUE)
+dataDir <- args[1]
+refDir <- args[2]
+
+
+gdsFile <-paste0(dataDir, "/2_gds/raw.gds")
+qcOutFolder<-file.path(dataDir, "/2_gds/QCmetrics")
+normgdsFile<-sub("\\.gds", "Norm.gds", gdsFile)
+autoData <-paste0(dataDir, "/3_normalised/autoOnlyNormBeta.rdata")
+configFile <- paste0(dataDir, "/config.r")
+epic2Manifest <- paste0(refDir,"/EPICArray/EPIC-8v2-0_A1.csv")
+
+
+source(configFile)
+
+arrayType <- toupper(arrayType)
+
+
+#----------------------------------------------------------------------#
+# LOAD PACKAGES
+#----------------------------------------------------------------------#
+library(bigmelon)
+library(data.table)
+
+
+#----------------------------------------------------------------------#
+# IMPORT DATA
+#----------------------------------------------------------------------#
+setwd(dataDir)
+
+normfile<-openfn.gds(normgdsFile, readonly = FALSE)
+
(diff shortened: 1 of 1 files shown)


array/DNAm/preprocessing/filterToAutosomes.r
```diff
@@ -0,0 +1,110 @@
...
+#----------------------------------------------------------------------#
+
+args<-commandArgs(trailingOnly = TRUE)
+dataDir <- args[1]
+refDir <- args[2]
+
+
+gdsFile <-paste0(dataDir, "/2_gds/raw.gds")
+qcOutFolder<-file.path(dataDir, "/2_gds/QCmetrics")
+normgdsFile<-sub("\\.gds", "Norm.gds", gdsFile)
```
sof202, line 28: Why use `sub` when we can just use: `paste0(dataDir, "/2_gs/rawNorm.gds")`?

This way each line would be consistent with one another. The `sub` isn't really adding anything except extra complexity.

array/DNAm/preprocessing/filterToAutosomes.r
```diff
@@ -0,0 +1,110 @@
...
+}
+
+if(arrayType == "HM450K"){
+load(file.path(refDir, "450K_reference/AllProbeIlluminaAnno.Rdata"))
+manifest<-probeAnnot[match(fData(normfile)$Probe_ID, probeAnnot$ILMNID), c("CHR", "INFINIUM_DESIGN_TYPE")]
+colnames(manifest) <- c("CHR", "Infinium_Design_Type")
+manifest$CHR <- paste0("chr", manifest$CHR)
+print("loaded hm450k manifest")
+rm(probeAnnot)
+}
```
sof202, line 85: Doesn't need to be done in this PR of course, just wanted to flag it. This is copied and pasted from another script. Having a function that is shared across files would be nice in the future. Especially considering how the column names of these manifest files seem to change for no reason between versions of manifests.

array/DNAm/preprocessing/filterToAutosomes.r
```diff
@@ -0,0 +1,110 @@
...
+
+
+#----------------------------------------------------------------------#
+# FILTER TO AUTOSOMAL PROBES ONLY
+#----------------------------------------------------------------------#
+
+if(arrayType == "V2" | arrayType == "HM450K"){
+    auto.probes<-which(manifest$CHR != "chrX" & manifest$CHR != "chrY")
```
sof202, line 93: Am I correct in assuming that chrM (mitochondrial) is not on the manifest files? If so, this is fine.

sof202 (approved):
Yeah this does what you want it to do. I'm surprised it takes 110 lines of code to accomplish this, definitely room for refactoring in the future considering the repeated code in other files in the pipeline.

How is this script going to be integrated into the pipeline? Is it just meant to ran separately by the user if they just want to view autosomes?

array/DNAm/preprocessing/filterToAutosomes.r
ew367: agreed. Logged as an issue for future

array/DNAm/preprocessing/filterToAutosomes.r
ew367: Nope, M is in there too. I'm not sure that they necessarily want to be removed though... I'm not aware of it being standard to remove them in the same way as it is the sex chromosome probes. Though I agree that leaving them in auto.probes is potentially misleading..

@ejh243 do you have an opinion?

ew367: It lookes like the M probes are actually excluded as part of bigmelons idat -> gds process right at the beginning of the pipeline, so while they are in the manifest, they are not in the data itself.

Issue #229, opened 2024-10-01
[Code Refactor]: only run cortical clock for brain samples


ew367:
### Path to affected file

```shell
calcQCmetrics.r
```

### Type of refactor

Code optimisation

### Description of required code refactor

The code currently runs the coritcal clock age predictor regardless of sample type, but it is only applicable to brain data. 
It could be turned off/on using config input ( `if (tissueType == "Brain")`  )

PR #236, opened 2024-10-25
Cortical clock check

sof202:
# Description

This pull request will add a check for the value of `tissueType` before executing the code block for the cortical clock.

There was also a lot of random indentation in this file, so I cleaned that up for everyone (some lines were indented for no reason, others had a mix of tabs and spaces).

## Problem

In QC.rmd we have the following lines where the `CCDNAmAge` field is expected to be in `QCMetrics`:
https://github.com/ejh243/BrainFANS/blob/6bbfbeceb89f79421ad8e94d1c47bf00ba2010ee/array/DNAm/preprocessing/QC.rmd#L1042-L1044

I'm not sure on how to proceed. I could make it so that this code also only runs when `tissueType == "brain"`, but I can't be sure if this is the desired behaviour. 

There are a few other instances of `CCDNAmAge` being used in QC.rmd. If the cortical clock is not ran, the required field in `QCMetrics` will not exist and this will fail. 

## Issue ticket number

This pull request is to address issue: #229.

## Type of pull request

- [ ] Bug fix
- [x] New feature/enhancement
- [ ] Code refactor
- [ ] Documentation update

## Checklist

### DIFF
--- a/array/DNAm/preprocessing/QC.rmd
+++ b/array/DNAm/preprocessing/QC.rmd
@@ -1035,13 +1035,13 @@ if (ageExists) {
 ```{r dnamage, results = 'asis', echo = FALSE, eval = ageValid}
 cat(
 "## Age Prediction\n",
-"The age of samples can be predicted from the DNAm data using the Epigenetic
-Clock algorithm developed by Steve Horvath. These predicted values are compared
-to the samples' reported ages. As on a sample level this estimaion can be
-inaccurate it is used as a quality check of the overall data set and not as a
-reason to exclude individual samples. The overall correlation was ",
+"The age of samples can be predicted from the DNAm data using the Epigenetic",
+"Clock algorithm developed by Steve Horvath. These predicted values are compared",
+"to the samples' reported ages. As on a sample level this estimaion can be",
+"inaccurate it is used as a quality check of the overall data set and not as a",
+"reason to exclude individual samples. The overall correlation was",
 signif(cor(as.numeric(QCmetrics$Age), QCmetrics$DNAmAge, use = 'pairwise.complete.obs'),3),
-"and the root mean square error was ",
+"and the root mean square error was",
 signif(sqrt(mean((QCmetrics$Age-QCmetrics$DNAmAge)^2, na.rm = TRUE)),3),
 "years.")
 ```
@@ -1059,9 +1059,9 @@ abline(a = 0, b = 1)
 
 ```{r ccdnamage, results = 'asis', echo = FALSE, eval = CCDNAmAgeExists}
 cat(
-"We also applied a custom clock algorithm developed secifically for cortical samples. The overall correlation was ",
+"We also applied a custom clock algorithm developed secifically for cortical samples. The overall correlation was",
 signif(cor(as.numeric(QCmetrics$Age), QCmetrics$CCDNAmAge, use = "pairwise.complete.obs"),3),
-"and the root mean square error was ",
+"and the root mean square error was",
 signif(sqrt(mean((QCmetrics$Age-QCmetrics$CCDNAmAge)^2, na.rm = TRUE)),3),
 "years.")
 ```
--- a/array/DNAm/preprocessing/QC.rmd
+++ b/array/DNAm/preprocessing/QC.rmd
@@ -1019,7 +1019,7 @@ if(length(errorIndexes) > 0){
 ```
 
 
-## Age Prediction
+<!-- Age Prediction -->
 
 ```{r ageSetup, echo=FALSE}
 ageExists <- "Age" %in% colnames(QCmetrics) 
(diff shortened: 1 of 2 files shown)


ew367:
I think the plot code wants to run if QCmetrics contains an age column AND a CCDNAmAge column, rather than just if tissue=brain.

The only other mention of CCDNAmAge I can see in the QC pipeline is line 1028 which is an inline mention. There is a wider error with this line however that if the QCmetrics object doesn't contain an 'Age' column, then you just get the text output with no values (e.g. " The overall correlation was and the root mean square error was years.")


One option would be to change that line to something like the following:

```{r checkAge, echo=FALSE}
# Check if "Age" exists and has non-NA values
ageExists <- "Age" %in% colnames(QCmetrics) && sum(!is.na(QCmetrics$Age)) > 0
CCDNAmAgeExists <- "CCDNAmAge" %in% colnames(QCmetrics) && ageExists

# Compute correlation and RMSE if conditions are met
corAge <- if (ageExists) signif(cor(as.numeric(QCmetrics$Age), QCmetrics$DNAmAge, use = "pairwise.complete.obs"), 3) else NA
rmseAge <- if (ageExists) signif(sqrt(mean((QCmetrics$Age - QCmetrics$DNAmAge)^2, na.rm = TRUE)), 3) else NA

corCCDNAmAge <- if (CCDNAmAgeExists) signif(cor(as.numeric(QCmetrics$Age), QCmetrics$CCDNAmAge, use = "pairwise.complete.obs"), 3) else NA
rmseCCDNAmAge <- if (CCDNAmAgeExists) signif(sqrt(mean((QCmetrics$Age - QCmetrics$CCDNAmAge)^2, na.rm = TRUE)), 3) else NA

```
The age of samples can be predicted from the DNAm data using the Epigenetic Clock algorithm developed by Steve Horvath. These predicted values are compared to the samples' reported ages. As on a sample level this estimation can be inaccurate it is used as a quality check of the overall data set and not as a reason to exclude individual samples. `r if (ageExists) "The overall correlation was "` `r if (ageExists) corAge` `r if (ageExists) " and the root mean square error was "` `r if (ageExists) rmseAge` `r if (ageExists) " years."`

`r if (CCDNAmAgeExists) "We also applied a custom clock algorithm developed specifically for cortical samples. The overall correlation was "` `r if (CCDNAmAgeExists) corCCDNAmAge` `r if (CCDNAmAgeExists) " and the root mean square error was "` `r if (CCDNAmAgeExists) rmseCCDNAmAge` `r if (CCDNAmAgeExists) " years."`

ew367 (commented):
Looks much tidier

sof202:
Ok I've made it so that the age related text and figure only appear if the age column exists and is valid. Then I've done the same for CCDNAmAge in a separate block. I'm confident the text will render as it did before, the only difference is that I changed the order of the plots to make more sense in the code. I can switch it back if needed. Also the plots will no longer be side by side as I cannot guarantee that both plots will be displayed without duplicating code. Again, if this is a requirement, I can change this back also.

array/DNAm/preprocessing/QC.rmd
```diff
@@ -1019,32 +1019,59 @@ if(length(errorIndexes) > 0){
...
-  title(main = paste("RMSE = ", signif(sqrt(mean(abs(QCmetrics$CCDNAmAge-QCmetrics$Age)^2, na.rm = TRUE)),3)), line = 2, adj = 1)
-  abline(model, lty = 2)
-  abline(a = 0, b = 1)
-}
+```{r dnamage, results = 'asis', echo = FALSE, eval = ageValid}
+cat(
+"## Age Prediction\n",
+"The age of samples can be predicted from the DNAm data using the Epigenetic",
+"Clock algorithm developed by Steve Horvath. These predicted values are compared",
+"to the samples' reported ages. As on a sample level this estimaion can be",
```
ew367, line 1040: should be 'estimation'

array/DNAm/preprocessing/QC.rmd
```diff
@@ -1019,32 +1019,59 @@ if(length(errorIndexes) > 0){
...
+"inaccurate it is used as a quality check of the overall data set and not as a",
+"reason to exclude individual samples. The overall correlation was",
+signif(cor(as.numeric(QCmetrics$Age), QCmetrics$DNAmAge, use = 'pairwise.complete.obs'),3),
+"and the root mean square error was",
+signif(sqrt(mean((QCmetrics$Age-QCmetrics$DNAmAge)^2, na.rm = TRUE)),3),
+"years.")
+```
+
+```{r dnamage-models echo=FALSE, fig.width = 10, fig.height = 5, eval = ageValid}
```
ew367, line 1049: missing comma after 'dnamage-models' here means that the report doesn't knit to html

ew367 (changes requested):
Missing comma on line 1049 needs to be fixed. And would be nice to fix the estimation typo at the same time. Otherwise, this works well.

ew367 (approved):

PR #248, opened 2024-11-08
Add conda environment for DNAm QC dependencies installation

sof202:
# Description

This pull request will throw away the old installLibraries.R script in favour of using a conda environment and renv environment. Currently, this is set up to work across the whole of the QC pipeline. I know that we said that we could split these environments up, I don't really see the point in having multiple renvs in the same directory. I can of course add new renvs to the analysis directory for example, but this is just a proof of concept for now.

## Testing

It would be great if lots of people can test this on both their local system and on the HPC (as there's a high likelihood it will break in some way on some system).

You don't need anything extra to check that this works than you normally do. Just get this branch and run the new setup script (then run the pipeline like normal)

```bash
cd path/to/BrainFANS
git fetch
git checkout -b add-conda-env origin/add-conda-env
array/DNAm/preprocessing/Setup/setup.sh .../path/to/your/config/file
```

You don't actually need a proper (fully filled out) config file for this to work. If you don't already have a config file for the DNAm QC pipeline you can just create a new text file somewhere and add the following:

```text
# Put the full file path to the root of the BrainFANS repository here:
SCRIPTSDIR=/path/to/BrainFANS
```

### If you don't have BrainFANS already

Then get BrainFANS with:

```bash
git clone https://github.com/ejh243/BrainFANS.git
```

Then follow the above.

## Issue ticket number

This pull request is to address issue: #166.

## Type of pull request

- [ ] Bug fix
- [x] New feature/enhancement
- [ ] Code refactor
- [ ] Documentation update

## Checklist

- [x] I have made the corresponding changes to the documentation


sof202:
Works for:

- [x] Luke (on ISCA)
- [x] Sam (on ISCA)

Does not work for:

marinafloresp:
Note that the path to the setup script is array/DNAm/preprocessing/Setup/setup.sh. I will try to see if it works for me

sof202:
> Note that the path to the setup script is array/DNAm/preprocessing/Setup/setup.sh. I will try to see if it works for me

Oh yeah. I moved it last second on Friday and forgot to update. Thank you.

array/DNAm/preprocessing/jobSubmission/1_runDNAmQC.sh
```diff
@@ -25,12 +25,18 @@ echo Log files intially stored in: ${SLURM_SUBMIT_DIR}/QCDNAdata_${SLURM_JOB_ID}
...
+fi
+    
+
 echo "Processing data located in :" ${DATADIR}
 
-## load modules
-echo "Loading R module :" $RVERS
-module load Pandoc
-module load $RVERS   # load specified R version
+source "${CONDA_SHELL}"
```
ejh243, line 38: Are these new shell variables that need to be included in the config template?

sof202: They don't need to be added to the config template. To make things easier for the user the setup script just adds these values for you.

sof202: I haven't removed `RVERS` from the example config file just yet as it is still being used in the other pipelines.

sof202:
The only problem I have with this PR right now is that it is currently installing packages that we might not need (as I just copied and pasted them from the previous script). This means that the following will always display in each error file:

```
- One or more packages recorded in the lockfile are not installed.
- Use `renv::status()` for more details.
```

I have tried a few ways to suppress these messages, but it doesn't seem likely to happen right now.

Also, this might look like an error, but it isn't. None of the packages aren't installed, instead there are 7 packages that are being installed, but are never explicitly used in the pipeline:


|package|     installed| recorded| used|
|---|-----|-----|------|
httr2    |   y   |      y       | n
R.methodsS3| y  |       y       | n
R.oo       | y  |       y       | n
R.utils    | y  |       y       | n
rentrez    | y  |       y       | n
rvest      | y  |       y       | n
selectr    | y  |       y       | n


I can remove these packages from the lock file if they aren't being used. But there is a chance that we do actually need them. For example, maybe a package we are using depends on one of these without actually saying so in the DESCRIPTION file. This would mean that if we removed said 'silent dependency' the pipeline would break.

rhaigh5 (approved):
All good! tested on BDR data and fully functional

Issue #198, opened 2024-09-11
[Bug]: clusterCellTypes.r script doesn't work when phenotype info is not available


ew367:
### What happened?

Line 165:166 of the clsuterCellTypes script defines columns to keep from QCmetrics dataframe. If the keepcols object is not longer than 1, line 167 creates a vector rather than a dataframe for object "uniqueIDs". Line 171 then cannot access "Individual_ID" by column name selection, as the uniqueIDs object has no columns.

### How can the bug be reproduced?

supply a samplesheet with no Age, Sex, Phenotype, Tissue.Centre columns

### Relevant log output

```shell
uniqueIDs$Individual_ID <- as.character(uniqueIDs$Individual_ID)
Error in uniqueIDs$Individual_ID :
  $ operator is invalid for atomic vectors
```

sof202:
So do we want to force `uniqueIDs` to be a data frame with `as.data.frame` then?

PR #263, opened 2024-11-25
Try to fix issue 198 - if uniqueIDs is a character, convert to a data frame

bethan-mallabar-rimmer:
# Description

In clusterCellTypes.r under 'Calculate individual FACS score' check if uniqueIDs is a character vector - if so convert it to a data frame. This stops uniqueIDs$Individual_ID <- as.character(uniqueIDs$Individual_ID) from breaking as $Individual_ID won't work on a character vector.

This pull request is to address issue: #198 

## Type of pull request

- [x] Bug fix
- [ ] New feature/enhancement
- [ ] Code refactor
- [ ] Documentation update

## Checklist

- [x] I have implemented fail safes in my code to account for edge cases _(note: it only converts uniqueID$Individual_ID to a dataframe if uniqueID is a character, which would be caused by QCmetrics lacking the Age, Sex, Phenotype, and Tissue.Centre columns. The code also works if QCmetrics contains no columns whatsoever. **The code breaks if QCmetrics is a vector [edited] - as I haven't used BrainFANS can someone let me know if this would ever actually happen?)**_

### DIFF
--- a/array/DNAm/preprocessing/clusterCellTypes.r
+++ b/array/DNAm/preprocessing/clusterCellTypes.r
@@ -164,9 +164,9 @@ for(i in 1:nrow(QCmetrics)){
 keepCols<-c("Individual_ID", "Sex", "Age", "Phenotype", "Tissue.Centre") 
 keepCols<-keepCols[keepCols %in% colnames(QCmetrics)]
 uniqueIDs<-unique(QCmetrics[,keepCols]) 
-if (is.character(uniqueIDs)) {
+if (!is.data.frame(uniqueIDs)) {
     uniqueIDs <- data.frame(Individual_ID=uniqueIDs)
-} 
+}
 
 indFACSEff<-aggregate(maxSD[which(QCmetrics$Cell_Type != "Total")], by = 
 list(QCmetrics$Individual_ID[which(QCmetrics$Cell_Type != "Total")]), FUN = median, na.rm = 
--- a/array/DNAm/preprocessing/clusterCellTypes.r
+++ b/array/DNAm/preprocessing/clusterCellTypes.r
@@ -161,16 +161,19 @@ for(i in 1:nrow(QCmetrics)){
 #----------------------------------------------------------------------#
 # CALCULATE INDIVIDUAL FACS SCORE
 #----------------------------------------------------------------------#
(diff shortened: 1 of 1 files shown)


sof202:
I guess to avoid the problem:

> The code breaks if QCmetrics contains only the Individual_ID columns and no other column...

We could just use `if (!is.data.frame(object)) { ...` instead of checking a specific type.

bethan-mallabar-rimmer:
> I guess to avoid the problem:
> 
> > The code breaks if QCmetrics contains only the Individual_ID columns and no other column...
> 
> We could just use `if (!is.data.frame(object)) { ...` instead of checking a specific type.

Actually nevermind, the code does work if you input a data frame with the Individual_ID column and no other column.
It's just that when you have a data frame with only one column, R has a tendency to convert it to a vector, and this is what caused it to break on my end. If QCmetrics is a vector, then the line `keepCols<-keepCols[keepCols %in% colnames(QCmetrics)]` results in the error `Error in QCmetrics[, keepCols] : incorrect number of dimensions`

If QCmetrics is generated earlier in the BrainFANS pipeline, I'm guessing the chances this object would be a vector rather than a data frame are quite slim?

With that said I prefer your suggestion `if (!is.data.frame(object)) { ...`  because it accounts for other classes that we might want to convert to a dataframe e.g. numeric vectors

ew367 (changes requested):
Requesting that this be updated to `if (!is.data.frame(object)) { ...` as suggested by @sof202

marinafloresp:
I agree with comments above. Checking first that the input QCmetrics is a dataframe will avoid later errors.

bethan-mallabar-rimmer:
> Requesting that this be updated to `if (!is.data.frame(object)) { ...` as suggested by @sof202


> I agree with comments above. Checking first that the input QCmetrics is a dataframe will avoid later errors.

I've now made this change

marinafloresp (approved):

bethan-mallabar-rimmer:
@ejh243 Are you able to confirm on your end whether the change I've made addresses your request please, so this pull can be merged and the issue closed? I'm unable to confirm on my end so just guessing that you may be able to. Thanks!

Issue #271, opened 2024-12-06
[Enhancement suggestion]: Use DT::datatables in DNAm QC rmarkdown report


sof202:
### Enhancement suggestion

@SziKayLeung has implemented datatables into their QC pipeline for ONT and it was agreed today that the rendering of this object type in rmarkdown reports is better than standard data.frames. We say this because if you have a lot of samples (lots of rows) a data.frame will take up a lot of space on the report (which can be annoying). However with datatables, the object will be rendered like this:

![image](https://github.com/user-attachments/assets/7e266df1-b8e4-46f4-a436-e466fcfe8597)

Which is much cleaner.

If the user that takes this on has a `QC.html` file that has been generated recently, this will help with identifying the lines of code behind each `data.frame` call that needs to be changed.

### Note:
If you don't have access to a rendered QC report, this issue can be probably still be completed by replacing all `data.frame` calls with `DT::datatable` (as the two are very similar for the most part). There are a few quirks I have seen in the past with this, however, someone can always check if your changes work.

PR #285, opened 2024-12-20
Use DT datatables for rendering tables in DNAm QC markdown report

sof202:
# Description

This pull request will, like the title suggests, use `datatables` from the `DT` package to render any tables that are in the DNAm QC rmarkdown report. The advantage of rendering tables this way is:
- Tables are now paginated
- Tables have a search feature
- Columns can be sorted (alphabetically or numerically)

There is an additional feature I have added here where a user has the option to save each table directly to a csv file with the click of a button (or just copy the data into the clipboard).

image

I have updated #248 to include the new `DT` package also.

## Question

With lines like:

```R
t(as.data.frame(table(table(QCmetrics$Individual_ID)))),
```

Do people think this is actually readable? It took me a good 20 minutes to figure out what this is actually doing.

Would anyone oppose to replacing this with something more verbose?

```R
# Better variables names can be chosen of course
id_counts <- table(QCmetrics$Individual_ID)
id_frequency_counts <- table(id_counts)
id_frequency_table <- t(as.data.frame(id_frequency_counts)
```

## Issue ticket number

This pull request is to address issue: #271.

## Type of pull request

- [ ] Bug fix
- [x] New feature/enhancement
- [ ] Code refactor
- [ ] Documentation update

## Checklist

### DIFF
--- a/array/DNAm/preprocessing/QC.rmd
+++ b/array/DNAm/preprocessing/QC.rmd
@@ -197,7 +197,7 @@ for(item in projVar){
   DT::datatable(
     as.data.frame(table(QCmetrics[, item])),
     caption = paste("Summary of samples categorised by", item),
-    col.names = c("Variable", "Number of Samples")
+    colnames = c("Variable", "Number of Samples")
   )
 }
 ```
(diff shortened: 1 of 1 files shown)


array/DNAm/preprocessing/QC.rmd
```diff
@@ -166,6 +161,12 @@ Data was loaded for `r nrow(QCmetrics)` samples from `r length(unique(QCmetrics$
 
 These are only included if they are categorical. For this study the following project variables were included:
 
+```{r, echo=FALSE,include = FALSE}
+# You need this code to conduct the magic dependences attaching so that we
+# can create DT::datatables() in a loop. I hate pandoc.
```
ejh243, line 166: Happy with the change of tool, perhaps remove the hate?

ejh243 (approved):
Do we need to document a general preference somewhere to use this approach over pandoc?

array/DNAm/preprocessing/QC.rmd
sof202: fair

PR #322, opened 2025-09-05
Update ATAC-seq pipeline

marinafloresp:
# Description

This pull request will update the ATAC-seq pipeline with the latest working scripts. Documentation still needs to be fully updated.

## Issue ticket number

This pull request is to address issue: #number.

## Type of pull request

- [ ] Bug fix
- [X] New feature/enhancement
- [ ] Code refactor
- [X] Documentation update

## Checklist

- [ ] I have performed a self-review of my own code


sof202 (approved):
Some nice changes here Marina. Some of the explanations in the Rmarkdown have much more clarity and succinctness than before.

This was pretty hard to review however. All I see on my end is one lump sum of all of your changes from the last couple of months. It would be easier to see your intentions behind changes if each change came as a different commit. As such I must assume I have missed a lot of minor details here. For example, you might remove a function somehwere, but I have no idea why it is has been removed (perhaps it has been replaced, moved or simply isn't needed anymore, I have no way of knowing).

I trust that you and Philippa have already been using these scripts and so they should be functional.

PR #353, opened 2026-06-08
build: Update renv environment

sof202:
# Description

This pull request will update the renv environment to once again work (bioconductor really doesn't like 'legacy'). Also uses renv 1.2.3, so package installation is now parallelised (making this a performance based PR too)


sof202:
Lint failing because superlinter hasn't been updated in 2 years I presume (trying to use a linter it doesn't have access to)

ew367 (commented):
Worked as expected for clean install followed by return of the entire pipeline

ew367 (approved):
Works as expected

